PR #1, opened 2021-12-12
Adding sweeps

KennethEnevoldsen:


sweep_config.json
```diff
@@ -0,0 +1,29 @@
...
+                128,
+                256,
+                512
+            ]
+        },
+        "dropout": {
+            "values": [
+                0.3,
+                0.4,
+                0.5
```
KennethEnevoldsen, line 25: Very high dropout, actually seems like deBERTa v. 2 removed the dropout during pertaining and add it in during fine-tuning.

train.py
```diff
@@ -0,0 +1,22 @@
...
+Inspiration: https://colab.research.google.com/github/wandb/examples/blob/master/colabs/pytorch/Organizing_Hyperparameter_Sweeps_in_PyTorch_with_W%26B.ipynb#scrollTo=r4VjKui20N3j
+"""
+
+from utils import read_json
+import wandb
+
+sweeps = read_json("sweep_config.json")
+
+
+def train(config=None):
```
KennethEnevoldsen, line 11: I think it is better to use the HF trainer instead. Here we can e.g. use 16 bit precision (as opposed the 32 or 64) or use the deepspeed library. I assume there is also other optimisations.

KennethEnevoldsen (commented):
Hi @MalteHB created this PR just to add some comments to the code, I know it is just and outline.

I would add it all into a `src` folder though, following the cookiecutter data science format.

MalteHB:
I agree with all of your comments, however, it was only a brief outline and never meant to be a PR.

PR #2, opened 2021-12-12
Adding datasets script

KennethEnevoldsen:
Add datasets script for converting datasets to HF


KennethEnevoldsen:
@HLasse I have added the initial version for DAGW can you update it once it is done?

HLasse:
Updated. The script is ready for HF hub, I just need to finalize the readme before pushing it there

src/data/danish-gigaword.py
```diff
@@ -0,0 +1,237 @@
...
+    def _info(self):
+        return datasets.DatasetInfo(
+            # This is the description that will appear on the datasets page.
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "text": datasets.Value("string"),
+                    "source": datasets.Value("string"),
+                    "doc_id": datasets.Value("string"),
+                    "LICENSE": datasets.Value("string"),
```
KennethEnevoldsen, line 101: det her kan vel sagtens være en kategorisk? Så at teksten ikke fylder for hvert felt?

src/data/danish-gigaword.py
```diff
@@ -0,0 +1,237 @@
...
+    ]
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            # This is the description that will appear on the datasets page.
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "text": datasets.Value("string"),
+                    "source": datasets.Value("string"),
```
KennethEnevoldsen, line 99: Categorical?

KennethEnevoldsen: Should there be an explanation attached to these variables?

src/data/danish-gigaword.py
```diff
@@ -0,0 +1,237 @@
...
+                    "doc_id": datasets.Value("string"),
+                    "LICENSE": datasets.Value("string"),
+                    "uri": datasets.Value("string"),
```
KennethEnevoldsen, line 102: What is uri?

src/data/danish-gigaword.py
```diff
@@ -0,0 +1,237 @@
...
+                    }
+
+                return metadict
+
+            # for twitter corpora
+            else:
+                return {
+                    "da_all_150420-260520.txt": {"uri": "NA", "date_built": "NA"},
+                    "da_fv19.txt": {"uri": "NA", "date_built": "NA"},
+                }
```
KennethEnevoldsen, line 218: seems odd I would instead do if section in ["datwitter...

Does this mean that there will be two NA rows? (which isn't clearly NA because it is strings?)Should it just not be included? Did we test if it was possible with an API key?

PR #12, opened 2021-12-20
Tokenisers

saattrupdan:
Two scripts related to tokenisers:

- `tokeniser_config.py`, containing the `TokeniserConfig` object for storing the exact configuration file for our tokenisers, to ensure reproducibility. Can be saved to a `json` file with the `save` method and loaded with the `load` class method.
- `train_tokeniser.py`, containing a `train_tokeniser` function for, guess what, training a tokeniser. Takes in a `TokeniserConfig` object as input, or a dictionary which is converted to such an object. Saves both tokeniser and config upon completion.

I did not include the _full_ flexibility of the tokenisers (there are thousands of combinations), but included a decent chunk of such flexibility. It supports three kinds of tokenisers (BPE, WordPiece and Unigram), and you can enable SentencePiece for them, to ensure that the tokens can be decoded as well (in the case of BPE and Unigram), which we'd need for generator models.

I took the liberty to add some "sensible defaults":

- Maximum sequence length defaults to 512 tokens.
- By default use NFKC Unicode normalisation, which converts weird symbols (like the triple elipsis symbol "…") to something more sensible (like three full stops "...")
- Special tokens are by default `<pad>`, `<s>`, `</s>`, `<unk>` and `<mask>`.
- Default is to add the CLS and SEP tokens (we can adjust what these tokens are in the config, defaults to `<s>` and `</s>`, as is the new standard).
- Add both padding and truncation by default

Further required configurations are:

- `tokeniser_type`  (`str`), can be 'bpe', 'wordpiece' or 'unigram'
- `vocab_size` (`int`)
- `lower_case` (`bool`)
- `sentence_piece` (`bool`)
- `byte_level` (`bool`)
- `add_prefix_space` (`bool`), whether to add a space before all words, including the first one. Used by, e.g., RoBERTa.

The `tests` folder contains a `tokenisers_test.py`, with unit tests for both scripts, which all pass on my end (but you should of course check if they work on your end as well!). For this reason I also included a `pytest.ini` file in the root, containing the `pytest` configuration.

I could set up a CLI as well, but I'm not sure how useful that is, compared to simply calling `train_tokeniser` from a notebook or a script. If you feel like such a thing would be useful I'll add it 🙂


pytest.ini
```diff
@@ -0,0 +1,5 @@
+[metadata]
+description-file=README.md
+
+[pytest]
+addopts=--verbose --durations=10 --color=yes -s -vv
```
KennethEnevoldsen, line 5: Never seen this before, so slowest of 10 repeats. Coloring the output (why is this not default). Looks good. We can automize this with a workflow that test on PR.

MalteHB (changes requested):
Incredible work @saattrupdan! The speed at which you work is quite amazing. 

I do not know how I feel about the british spelling, and I actually think that we should change it to "tokenizer" with "z" for streamlinings sake. 

Any other thoughts on this? @HLasse @KennethEnevoldsen

pytest.ini
KennethEnevoldsen: Added it as an issue #13

src/tokeniser_config.py
```diff
@@ -0,0 +1,260 @@
...
+        self.bos_token = bos_token
+        self.eos_token = eos_token
+        self.unk_token = unk_token
+        self.mask_token = mask_token
+
+        # Lower case `tokeniser_type` if possible
+        try:
+            self.tokeniser_type = self.tokeniser_type.lower()
+        except AttributeError:
+            pass
```
KennethEnevoldsen, line 93:
```suggestion
        self.tokeniser_type = self.tokeniser_type.lower()
```
You are assuming the tokenizer type is string.

src/tokeniser_config.py
```diff
@@ -0,0 +1,260 @@
...
+    def check_config(self):
+        '''Check that the configuration is valid.
+
+        This both checks that the data types of the inputs are correct, as well
+        as the values.
+
+        Raises:
+            ValueError:
+                If the configuration is invalid.
+        '''
```
KennethEnevoldsen, line 107: Have you tried pydantic? HF and spaCy (among others) use it for their base classes.

src/train_tokeniser.py
```diff
@@ -0,0 +1,142 @@
...
+                                            show_progress=show_progress,
+                                            special_tokens=special_tokens)
+    elif config.tokeniser_type == 'unigram':
+        trainer = trainers.UnigramTrainer(vocab_size=config.vocab_size,
+                                          special_tokens=special_tokens,
+                                          show_progress=show_progress,
+                                          unk_token='<unk>')
+
+    # Train the tokeniser
+    tokeniser.train_from_iterator(iterator=dataset['text'],
```
KennethEnevoldsen, line 130: Does this work for streaming datasets:

```
ds = load_dataset("DDSC/angry-tweets", streaming = True)
ds = ds["train"]
ds["text"]
# TypeError: 'IterableDataset' object is not subscriptable
```

One option would be to just take in an iterator?

KennethEnevoldsen: then you could simply input:
```
(e["text"] for e in iterable_dataset)
```

KennethEnevoldsen (changes requested):
Looks good. The issue with the dataset seems to be the most problematic. I would also recommend taking a look at pydantic, seems like it would have solved a lot of the type checking for you.

saattrupdan:
> Incredible work @saattrupdan! The speed at which you work is quite amazing.
> 
> I do not know how I feel about the british spelling, and I actually think that we should change it to "tokenizer" with "z" for streamlinings sake.
> 
> Any other thoughts on this? @HLasse @KennethEnevoldsen

Thanks @MalteHB! Haha yeah, I guess I've been indoctrinated. We could all just do what the Americans are doing, sure. I'll change that ASAP.

src/train_tokeniser.py
saattrupdan: Good catch! I've implemented the iterator now, and included a unit test for streamed datasets.

src/tokeniser_config.py
saattrupdan: Simply removed lower casing now, with pydantic implemented. This means that you _have_ to write 'bpe' and not, say, 'BPE'. But I think we can handle that anyway 😛

src/tokeniser_config.py
saattrupdan: I had never heard about it, but it's very cool! Implemented it now.

KennethEnevoldsen: Well glad to give you some new tools ;)

src/tokeniser_config.py
KennethEnevoldsen: We wil manage ;)

src/train_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+                                            special_tokens=special_tokens)
+    elif config.tokenizer_type == 'unigram':
+        trainer = trainers.UnigramTrainer(vocab_size=config.vocab_size,
+                                          special_tokens=special_tokens,
+                                          show_progress=show_progress,
+                                          unk_token='<unk>')
+
+    # Train the tokenizer
+    dataset_iter = (sample['text'] for sample in dataset)
+    tokenizer.train_from_iterator(iterator=dataset_iter, trainer=trainer)
```
KennethEnevoldsen, line 131:
```suggestion
    # Train the tokenizer
    tokenizer.train_from_iterator(iterator=texts, trainer=trainer)
```

src/train_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+from pathlib import Path
+from typing import Union
+from .tokenizer_config import TokenizerConfig
+
+
+def train_tokenizer(dataset: Dataset,
+                    config: Union[TokenizerConfig, dict],
+                    save_tokenizer: bool = True,
+                    output_dir: Union[str, Path] = '.',
+                    show_progress: bool = True):
```
KennethEnevoldsen, line 15:
```suggestion
def train_tokenizer(texts: Iterable[str],
                    config: Union[TokenizerConfig, dict],
                    save_tokenizer: bool = True,
                    output_dir: Union[str, Path] = '.',
                    show_progress: bool = True):
```

src/train_tokenizer.py
```diff
@@ -0,0 +1,142 @@
+'''Script to train tokenizers from a Hugging Face dataset'''
+
+from tokenizers import (trainers, pre_tokenizers, tokenizers, normalizers,
+                        models, AddedToken, processors, decoders)
+from datasets.arrow_dataset import Dataset
+from pathlib import Path
+from typing import Union
```
KennethEnevoldsen, line 7:
```suggestion
from typing import Union, Iterable
```

src/train_tokenizer.py
```diff
@@ -0,0 +1,142 @@
...
+                    config: Union[TokenizerConfig, dict],
+                    save_tokenizer: bool = True,
+                    output_dir: Union[str, Path] = '.',
+                    show_progress: bool = True):
+    '''Train a tokenizer on a dataset.
+
+    Args:
+        dataset (Dataset):
+            Dataset to train the tokenizer on. Must have a feature named
+            'text'.
```
KennethEnevoldsen, line 21:
```suggestion
        texts (Iterable[str]): An iterable of strings to train the tokenizer on.
```

KennethEnevoldsen (changes requested):
Looks good. Only would want to change the dataset input to an iterable of strings. I will set the tokenizer to train tomorrow.

src/train_tokenizer.py
saattrupdan: Fair enough, that would make it more versatile for sure. I won't accept your suggestion here, as the unit tests will fail. For maximum versatility, I'll set it up to accept either a HF Dataset, or an iterable of texts.

saattrupdan: I have now changed the input to accept either a `Dataset`, an `IterableDataset` or an iterable of strings. Also included a few more unit tests to test these, and they all run successfully.

KennethEnevoldsen: Perfect, yea this is probably the best way to go.

Issue #13, opened 2021-12-21
Add testing workflow


KennethEnevoldsen:
Add git workflow for pytests.

KennethEnevoldsen:
Merged in #24

PR #24, opened 2021-12-23
Added pytest workflow for automatic testing

KennethEnevoldsen:
Added pytest workflow pr. #13 .

This workflow runs the tests on Python 3.7, 3.8, 3.9. On the latest versions of Mac, Linux, and Windows. It also generates a comment to which contains the code coverage.


litterature.md
```diff
@@ -1,4 +1,4 @@
-# Litererature
+# Littererature
```
saattrupdan, line 1: Typo: Literature

tests/datasets_test.py
```diff
@@ -38,6 +41,7 @@ def test_load_reddit():
     assert isinstance(ds[0], dict)
     assert "text" in ds[0]
```
saattrupdan, line 43: PEP8: Needs two empty lines between global functions.

tests/datasets_test.py
```diff
@@ -12,6 +13,7 @@ def test_load_tweets():
     assert isinstance(tweet[0], dict)
     assert "text" in tweet[0]
```
saattrupdan, line 15: PEP8: Needs two empty lines between global functions.

litterature.md
```diff
@@ -1,4 +1,4 @@
...
 This is a working document for literature. This is much for sharing and organizing meterial as opposed to the Zotero which is for citation management.
```
saattrupdan, line 2: Typo: material

tests/datasets_test.py
```diff
@@ -4,6 +4,7 @@
 from dfm.data.load import load_tweets, load_news, load_dagw, load_reddit, load_tokenizer_ds
 from datasets import IterableDataset, Dataset
 
+@pytest.mark.skip(reason="Dataset not publicly available")
```
saattrupdan, line 7: PEP8: Needs two empty lines before global functions.

saattrupdan (changes requested):
I'm guessing the files in the `dev` folder are auto-generated?

KennethEnevoldsen:
No the files in dev are manually added by me (copy-paste from other projects). You could essentially wrap it in a GitHub action and remove it from the codebase (which I suppose would be ideal, but would have to spend to time looking into how to write a GitHub action).

Corrected the typos and formatted them to Black. Also added a PR to check if the code style is black. We can use another style as well but had the workflow lying around for black.

requirements.txt
```diff
@@ -15,4 +15,7 @@ wasabi~=0.9.0
 pydantic~=1.8.2
 
 # testing
-pytest~=6.2.5
\ No newline at end of file
+pytest~=6.2.5
+
+# format
+black
```
saattrupdan, line 21: Version?

saattrupdan (changes requested):

requirements.txt
KennethEnevoldsen: Now added

KennethEnevoldsen:
@HLasse will you take a look at this when you are back from vacation? Seems like there is an issue with loading DAGW from HF hub.

This has previously been an issue with the name change, but that does not seem to be the case now. An alternative might be that it is simply not up either on HF (which I doubt) or on DAGW (which I believe is more likely).

PR #29, opened 2022-01-10
Lex.dk

saattrupdan:
Tiny script that loads the Lex.dk from a JSONL file.

I set the `date` feature to simply have the `string` data type, as the time zones in the dataset are variable, and PyArrow only supports dates with the same timezone. The dates can easily be converted to `datetime` objects using the `datetime.fromisoformat` function.


dfm/data/lexdk/lexdk.py
```diff
@@ -0,0 +1,31 @@
...
+        authors=[Value('string')],
+        date=Value('string'),
+        text=Value('string')
+    ))
+    return Dataset.from_json(path, features=features)
+
+
+if __name__ == '__main__':
+    dataset = load_lexdk('lexdk.jsonl')
+    breakpoint()
```
KennethEnevoldsen, line 31: Maybe just add this as a test? But without running it using:
```
@pytest.mark.skip(reason="Dataset not publicly available")
```

KennethEnevoldsen (commented):
Looks good, would add a test as well, otherwise, nothing else to add.

MalteHB:
Looks great @saattrupdan! And nice addition with the test. It seems that it is not black formatted?

dfm/data/lexdk/lexdk.py
saattrupdan: Added a test now 👌

saattrupdan:
Added a test now, moved the loading script to `load.py` (which looks for the dataset in the `dfm/data/lexdk` folder), and updated the number of tokens in the dataset in the readme.

saattrupdan:
> Looks great @saattrupdan! And nice addition with the test. It seems that it is not black formatted?

Fixed now 👌

MalteHB (commented):
Good job!

PR #44, opened 2022-01-29
Feature/dedup

saattrupdan:
Implementation of deduplication of a dataset using the MinHash algorithm.

Can work on the following levels:

- Entire documents (set `split_method=None`)
- Paragraphs (set `split_method='paragraph'`)
- Word n-grams (set `split_method='word_ngram'` adjust `ngram_size` and `ngram_stride` as necessary)
- Character n-grams (set `split_method='char_ngram'` adjust `ngram_size` and `ngram_stride` as necessary)

It streams in the dataset, and streams the deduplicated data into a JSONL file. Only the MinHash cache will be held in memory, which is really minimal.

Did not implement any multiprocessing, as we're streaming one sample at a time anyway, so not sure if this is something we can do. We also need to compare the current document to previous ones, which will be quite messy in a multiprocessing setup.

Sample usage:

```python
from datasets import load_dataset
corpus = load_dataset('DDSC/partial-danish-gigaword-no-twitter', streaming=True, split='train')
deduper = Deduper(split_method='paragraph')
deduper.deduplicate(corpus, output_fname='deduplicated.jsonl')
```

Currently missing unit tests, but these could be added on the next hackathon (or before).


dfm/data/deduper.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        # Convert corpus to an iterable of strings if a Dataset is given
+        if isinstance(corpus, Dataset) or isinstance(corpus, IterableDataset):
+            corpus = (sample["text"] for sample in corpus)
+
+        # Ensure that `output_fname` is a Path object
+        output_fname = Path(output_fname)
+
+        # If the output file already exists then raise an error
+        if output_fname.exists():
```
MalteHB, line 181: You could consider including an `override` keyword argument, defaulting to `False`, instead of having to rename the output file.

MalteHB (approved):
Looks really good @saattrupdan! Again - I love your style of coding. It is really intuitive to read, and easily comprehendible. Good job!

dfm/data/deduper.py
saattrupdan: Good idea. Added now 😊

dfm/data/deduper.py
```diff
@@ -0,0 +1,233 @@
...
+                doc[i : i + self.ngram_size]
+                for i in range(0, max_char_idx, self.ngram_stride)
+            ]
+        elif self.split_method == "word_ngram":
+            words = doc.split(" ")
+            max_word_idx = len(words) - self.ngram_size
+            shingles = [
+                " ".join(words[i : i + self.ngram_size]).strip()
+                for i in range(0, max_word_idx, self.ngram_stride)
+            ]
```
KennethEnevoldsen, line 117: We should probably filter out empty strings here, so:

```suggestion
            shingles = list(filter(lambda x: x,
                (" ".join(words[i : i + self.ngram_size]).strip()
                for i in range(0, max_word_idx, self.ngram_stride))
            ))
            
```

Screenshot 2022-01-29 at 21 57 05

KennethEnevoldsen:
We should also remove the text-dedup dependency as well as the dedupe.py in the data folder. 

I have also created a cleaning folder to differentiate between data loaders and data cleaners.

KennethEnevoldsen (changes requested):
The filter change is probably important, otherwise, it seems good.

dfm/data/deduper.py
saattrupdan: Good point. I tried to keep things consistent and fixed it in the definition of `words` instead.

dfm/data/deduper.py
```diff
@@ -0,0 +1,233 @@
...
+Author:
+    Dan Saattrup Nielsen (saattrupdan@gmail.com)
+
+References:
+    [1] Broder, Andrei Z. "On the resemblance and containment of documents."
+        Proceedings. Compression and Complexity of SEQUENCES 1997
+        (Cat. No. 97TB100171). IEEE, 1997.
+"""
+
+from datasketch import MinHash, LeanMinHash, MinHashLSH
```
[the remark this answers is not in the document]
saattrupdan: Yes! Included it in requirements now.

saattrupdan:
@KennethEnevoldsen 

> We should also remove the text-dedup dependency as well as the dedupe.py in the data folder.

This should already be done?

> I have also created a cleaning folder to differentiate between data loaders and data cleaners.

I can't see this folder. It's not on the main branch?

mollerhoj:
Looks very good!

KennethEnevoldsen (approved):

PR #48, opened 2022-02-01
Allow usage of hashing trick instead of shingles

mollerhoj:
Depends on https://github.com/centre-for-humanities-computing/danish-foundation-models/pull/47

This allows the deduper to use the hashing trick instead of shingles.

Mention here (from https://arxiv.org/pdf/2201.11990.pdf)
```
Exact match duplicates would be computationally expensive, so we opted to take a fuzzy deduplication
approach similar to other works [9, 17]. We used a hashing vectorizer with 1,048,576 features to vectorize
documents (HashingVectorizer from scikit-learn4
), calculated min-hashes of the vectorized documents
(using datasketch5
), and performed Locality Sensitive Hashing (LSH) through datasketch on all minhashes in order to identify potential duplicates. We set our LSH parameters in such a way as to increase the
likelihood that documents with Jaccard similarity ≥ 0.8 would occur in at least one LSH bucket together.
Specifically, we used 20 bands of size 13 for a total of 260 hash functions.
```

Do note that I don't see why using the hashing trick here gives any advantage? Perhaps the implementation in the paper uses multiple passes and thus have to store the vectors? Given that this only uses a single pass, I don't see why it would improve anything. @saattrupdan what do you think?

### DIFF
--- a/dfm/cleaning/deduper.py
+++ b/dfm/cleaning/deduper.py
@@ -21,6 +21,7 @@ from unicodedata import normalize
 import re
 from tqdm.auto import tqdm
 from collections.abc import Callable
+from sklearn.feature_extraction import FeatureHasher
 
 
 def _default_normalization(doc: str):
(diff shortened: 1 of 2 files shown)


saattrupdan:
@mollerhoj Maybe the main benefit of the hashing trick is to use less memory? I mean, in case there are a gazillion shingles in a large document, maybe feature hashing will just use less memory while building the minhash for it? That's the only benefit I can see at least. I don't see anything wrong with adding it though, so this looks fine to me. But let's sort out the unit test branch first 🙂

mollerhoj:
> 

Yes, thats what I thought. But given that our deduper uses a single pass, saving memory here is silly. This could run on a raspberry pie without using too much memory (not taking the minhash cache into account, but thats unaffected by this change).

In other words, the only benefit of this is to replicate the paper - I see 0 added value in how they do it - it only makes the implementation a bit more obscure.

saattrupdan:
Closing the PR in that case. We can always reopen it if we need it.

PR #49, opened 2022-02-02
quality filter cleanup

mollerhoj:
This PR simply improves the code quality of the quality_filter unit tests, mainly by splitting assertions up into individual tests, and the fixture data into multiple sets of data rather than a long list.

No tests were harmed during this cleanup.

A suggestion for @KennethEnevoldsen and @KasperJuunge. Would it make sense to split this into one class per quality check? It seems to me that the checks are unrelated.


tests/cleaning/quality_filter_test.py
```diff
@@ -0,0 +1,218 @@
...
+        return """
+        [summary]
+
+        * test 1
+            * test 2
+            * test 3
+        * test 1
+            * test 2
+                * test 3
+        """
```
KennethEnevoldsen, line 59: This is also a bullet point text (ellipsis is the ...)

KennethEnevoldsen (changes requested):
Looks good to me. I would say with this rewrite you don't use the class at all so you might as well remove it. @saattrupdan is there any benefit of having the class which I am unaware of?

tests/cleaning/quality_filter_test.py
```diff
@@ -0,0 +1,218 @@
...
+            "jeg er glad",
+            "56789okd23456789098765sds",
+            "jeg er glad...",
+            "jeg er glad…",
+            "67 54 13 B7",
+            "#yolo # test ##",
+        ]
+
+    @pytest.fixture(scope="class")
+    def long_text(self):
```
KennethEnevoldsen, line 20:
```suggestion
    def valid_text(self):
```
Is probably more descriptive

tests/cleaning/quality_filter_test.py
```diff
@@ -0,0 +1,218 @@
...
+    def quality_filter(self):
+        return QualityFilter()
+
+    def test_good_stop_words(self, quality_filter, stop_words):
+        assert (
+            quality_filter.stop_word(
+                quality_filter.nlp("jeg er glad"), n=2, stop_words=stop_words
+            )
+            is True
+        )
```
KennethEnevoldsen, line 106: Might as well test with all short texts splitting them into valid/invalid

tests/cleaning/quality_filter_test.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_good_bullets(self, quality_filter):
+        assert (
+            quality_filter.line_bullets_or_ellipsis(
+                quality_filter.nlp("56789okd23456789098765sds"),
+                max_p_bullets=0.5,
+                max_p_ellipsis=1,
+            )
+            is True
+        )
```
KennethEnevoldsen, line 126: test on long text as well

tests/cleaning/quality_filter_test.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_bad_bullets_on_ellipsis(self, quality_filter, ellipsis_text):
+        assert (
+            quality_filter.line_bullets_or_ellipsis(
+                quality_filter.nlp(ellipsis_text), max_p_bullets=0.5, max_p_ellipsis=1
+            )
+            is False
+        )
```
KennethEnevoldsen, line 134: Again this is False because of the bullet point part, not the ellipsis part (i.e. this and the next test just test two different types of bullets)

tests/cleaning/quality_filter_test.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_bad_ellipsis_dots(self, quality_filter):
+        assert (
+            quality_filter.line_bullets_or_ellipsis(
+                quality_filter.nlp("jeg er glad..."),
+                max_p_bullets=1.0,
+                max_p_ellipsis=0.5,
+            )
+            is False
+        )
```
KennethEnevoldsen, line 160: This is basically a copy of the next test you can simplify it using `pytest.mark.parametrize `
```suggestion
    @pytest.mark.parametrize(
    "text,valid",
    [
        ("jeg er glad...", False),
        ("jeg er glad…", False),
    ],
    )
    def test_bad_ellipsis_dots(self, quality_filter, text, valid):
        assert (
            quality_filter.line_bullets_or_ellipsis(
                quality_filter.nlp("jeg er glad…"),
                max_p_bullets=1.0,
                max_p_ellipsis=0.5,
            )
            is valid
        )
```

KennethEnevoldsen (changes requested):
I read through it again, there are probably a few more clean up steps to do.

saattrupdan:
> Looks very good. I would say with this rewrite you don't use the class so you might as well remove it. @saattrupdan is there any benefit of having the class which I am unaware of?

@KennethEnevoldsen It's basically a matter of organisation, and there are multiple ways of doing that. It seems that we're sticking to the "one test script for each script", in which case it makes sense to be able to group the tests in the individual test scripts. As an added benefit, we can treat these groups (=classes) as one unit: disable all at once, use `pytest.parametrize` to add more test cases to all tests in a class, subclass etc.

KennethEnevoldsen:
Ahh perfect, thanks for the clarification. Yea so probably good to keep the test of the entire class in a class just to group it.

mollerhoj:
I generally agreed with this review. However do note that the scope of this PR is cleaning up *without* changing the semantics for the tests. Further improvements should be in separate subsequent PRs, so I will ignore those comments here.

KennethEnevoldsen:
I have merged this into #54, where I made some changes to the quality filter. I also updated the test with the above-mentioned. @mollerhoj I have added you as a reviewer to that one to ensure consistency.

PR #50, opened 2022-02-03
Fix/unit test deduper

saattrupdan:
This fixes a couple of smaller bugs in the unit-test-deduper, but mainly it reorganises all the tests using pytest parametrization. This is quite subjective, but if you feel like it's nicer then go ahead and accept the merge 🙂


mollerhoj:
I find some of these changes good, some bad.

Using idxs to refer to positions in an arbirary array is very bad for readability.

Tests serve 2 purposes: Increasing the confidence in correctness, and serve as a specification. This change improves the former but deteriorate the latter.

Can we leverage parameterization to increase the number of asserting without sacreficing readbility?

saattrupdan:
> I find some of these changes good, some bad.
> 
> Using idxs to refer to positions in an array is bad for readability.
> 
> Tests serve 2 purposes: Increasing the confidence in correctness, and serve as a specification. This change improves the former but deteriorates the latter.
> 
> Can we leverage parameterization to increase the number of assertions without sacreficing readbility?

That's fair enough. The only reason why I chose the indices was to not have tonnes of strings floating around in the code, but I'd be up for replacing the indices with the texts again. But I'm glad that you're happy about using the parametrisation, as it also enables plugging in more test cases easily 🙂

saattrupdan:
@mollerhoj Made the changes now.

mollerhoj:
@saattrupdan. Thanks. Looking at it again, I can't help but feel this it is still a regression regarding readability - But there is a good chance I'm just blinded by having specific preferences/being used to different setups. This code is clever - but I'm rather opposed to clever code. I'm just not sure if this is idiomatic or not as python is not my day2day language. Could some 3 person confirm that this change is indeed easier to read for them then I'd be happy to merge.

Regarding plugging in more test cases, I agree that I'll require less lines of code here of course. But what I'm missing the different names of the tests - tests should usually be named as to explain the purpose of the test, but that's not happing here due to parameterization.

In Haskell, we use parameterization to increase the number assertions in a single test case. In this PR, it is used to merge similar test cases into a single test.

You are free to merge of course if you disagree with these points of course.

mollerhoj (changes requested):
@saattrupdan. Thanks. Looking at it again, I can't help but feel this it is still a regression regarding readability - But there is a good chance I'm just blinded by having specific preferences/being used to different setups. This code is clever - but I'm rather opposed to clever code - Especially for tests. I'm just not sure if this is idiomatic or not as python is not my day2day language. Could some 3 person confirm that this change is indeed easier to read for them then I'd be happy to merge.

Regarding plugging in more test cases, I agree that I'll require less lines of code here of course. But what I'm missing the different names of the tests - tests should usually be named as to explain the purpose of the test, but that's not happing here due to parameterization. It could be done with comments on different parameter setups, but comments are not inforced, and they don't show up when the tests fail. (The test name will usually be an indicator of why it failed. With this change all you'll get back is that the 'deduplication test' failed, with does not tell you much. 

In Haskell, we use parameterization to increase the number assertions in a single test case. In this PR, it is used to merge similar test cases into a single test.

You are free to merge if you disagree with these points of course.

saattrupdan:
@mollerhoj To be honest, I don't think we should be spending this amount of time on a thing like unit test code styles. I'll close this PR and we'll stick to your setup, no worries.

PR #51, opened 2022-02-04
Profanity filter

mollerhoj:
Beginning of a very simple profanity filter.

https://arxiv.org/abs/2112.11446 uses googles "safe search" which is not applicable here.

Thoughts:
- I couldn't find a list of safe danish domains. I don't think we can just use an international one, because looking at the texts, I believe most of the adult material here come from:
- A) Small, rather fringe sites (highly SEO optimized sites that try to lure in customers by targeting specific sexual preferences).
- B) Escort sites offering their services in specific cities (Esbjerg was a very common word). My guess is that since prostitution is legal in Denmark, the Danish web has an overrepresentation of this kind of stuff that targets sex tourism.
- I couldn't find a comprehensive wordlist, so instead I found about 20 seed words, and then bootstrapped from there (manually adding words from the documents found by using this filter). Hence no source.
- Regarding specific biases, I believe the best way forward would be to actually classify the sites and train a classifier? The wordlist here is still very useful for manually finding pornographic sites for such a data set. A simpler approach would to just remove the troublesome words mentioned, or manually add a weight to them, and use that.
- Perhaps we could use the classifier to create a list of domains to block - I'm not sure how many candidates we would end up with, but maybe its small enough that its possible to manually curate it?

Todo:
- Choose a better heuristic
- Expand wordlist (include english list also)
- Run tests

On a site note: Do remember that all filtering adds extra bias to any data set. What might seem politically correct or in good taste is in no way a reflection of any objective truth - only the moral whims of the time. We should of course try to avoid false positives. If the goal is to build an AI that can be used in commercial settings, censoring on specific words makes sense because that the usual technique used by nsfw filters - The American television *biip* sound comes to mind.

PR #52, opened 2022-02-05
Create dataset-description values 

martbern:
Description of datasets. Gets a match-count for spaCy matcher objects. This allows us rather complicated rules: https://spacy.io/usage/rule-based-matching#phrasematcher

Implemented counts for occupations, gendered terms, religions and names.

Might want to extract some of the functions to a more general utility file or similar, if we want e.g. #51 to share the codebase.

In future PRs:
* Visualisations
* General benchmarking utility

Missing:
* Tests
* Visualisations


.gitignore
```diff
@@ -133,3 +133,5 @@ dmypy.json
 
 # vs code
 .vscode/*
+csv/output.csv
```
KennethEnevoldsen, line 136:
```suggestion

# Data files
*.csv
```

.gitignore
```diff
@@ -133,3 +133,5 @@ dmypy.json
...
+.DS_Store
```
KennethEnevoldsen, line 137:
```suggestion

# MacOS
.DS_Store
```

dfm/description/description_pattern_lists.py
```diff
@@ -0,0 +1,89 @@
+occupation_pattern_list = [
+    "revisor",
+    "administrator",
+    # "rådgiver",  # Matches many non-occupation tokens, maybe use lemmatizer?
```
KennethEnevoldsen, line 4: Lemmatizer would be too slow. Not an option.
```suggestion
    # "rådgiver",  # Matches many non-occupation tokens
```

dfm/description/description_pattern_lists.py
```diff
@@ -0,0 +1,89 @@
+occupation_pattern_list = [
```
KennethEnevoldsen, line 1: We should probably expand this list with pluralis og "bestemt ental"

KennethEnevoldsen: Probably goes for all the lists.

dfm/description/gen_description.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    Generates matcher objects from a list of dictionaries with {matcher_label (str): pattern (list)}
+    Pattern must conform to SpaCy pattern standards:
+
+    Example:
+        >>> pattern_container_list = [
+        >>>    {"atheism": [{"LOWER": {"REGEX": "athei.+"}}]},
+        >>>    {"atheism": [{"LOWER": {"REGEX": "atei.+"}}]},
+        >>>    {"skøde": [{"LOWER": "skøde"}]},
+        >>> ]
```
KennethEnevoldsen, line 85: Example is typically reserved for a code example using the function, not function output.

dfm/description/gen_description.py
```diff
@@ -0,0 +1,220 @@
...
+
+        counts[pattern_label] = 0
+
+    for match_id, start, end in matcher_object(doc):
+        counts[nlp.vocab.strings[match_id]] += 1
+
+    return dict(counts)
+
+
+def get_match_counts_from_batch(batch, matcher_object: Matcher, nlp: Language) -> dict:
```
KennethEnevoldsen, line 124: I would refactor so that this takes texts and create one specifically for HF batches.

dfm/description/gen_description.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    Takes a spacy batch of docs and processes them into a dictionary with
+    {match_label (str): match_counts (list of ints)}
+    """
```
KennethEnevoldsen, line 128: Is there a reason to have a list of match counts (vs. the sum?)

dfm/description/gen_description.py
```diff
@@ -0,0 +1,220 @@
...
+
+    for doc in docs:
+        doc_match_counts = get_match_counts_from_doc(doc, matcher_object, nlp)
+
+        for pattern_label in doc_match_counts.keys():
+            pattern_match_count = doc_match_counts.get(pattern_label, 0)
+
+            batch_match_counts[pattern_label].append(pattern_match_count)
+
+    return dict(batch_match_counts)
```
KennethEnevoldsen, line 142:
```suggestion
    return batch_match_counts
```

requirements.txt
```diff
@@ -12,7 +15,7 @@ spacy>=3.1.4,<=3.3.0
 datasketch~=1.5.6
 
 # utility
-wasabi~=0.9.0
+wasabi~=0.8.2
```
KennethEnevoldsen, line 18: Is there a problem with version 0.9.0?

KennethEnevoldsen (commented):
Looks good, a few things we probably need to add to this:
- [x] An approximation on speed (potentially test with dict if too slow)
- [x] Inclusion of the pornographic words from #51

dfm/description/gen_description.py
martbern: Yeah, it includes more information :-) 

We want the distribution of match counts across a corpus – so we want the match counts for each text.

dfm/description/gen_description.py
martbern: Added

requirements.txt
martbern: Conflict with dacy 1.3.0 - ERROR: dacy 1.3.0 has requirement wasabi<0.9.0,>=0.8.2, but you'll have wasabi 0.9.0 which is incompatible.

martbern:
> * [ ]  An approximation on speed (potentially test with dict if too slow)

176 seconds on 1% of DAGW on an M1 👍

dfm/description/description_pattern_lists.py
```diff
@@ -0,0 +1,436 @@
+occupation_pattern_list = set([
```
KennethEnevoldsen, line 1: You typically don't want data in scripts. I suggest writing these to .json files and read them in.

dfm/description/gen_description.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Generates matcher objects from a list of dictionaries with {matcher_label (str): pattern (list)}
+        Pattern must conform to SpaCy pattern standards:
+
+        Example input:
+            >>> pattern_container_list = [
+            >>>    {"atheism": [{"LOWER": {"REGEX": "athei.+"}}]},
+            >>>    {"atheism": [{"LOWER": {"REGEX": "atei.+"}}]},
+            >>>    {"skøde": [{"LOWER": "skøde"}]},
+            >>> ]
```
KennethEnevoldsen, line 94: The default way to write an example is using Example or Examples

```suggestion
        Example:
            >>> pattern_container_list = [
            >>>    {"atheism": [{"LOWER": {"REGEX": "athei.+"}}]},
            >>>    {"atheism": [{"LOWER": {"REGEX": "atei.+"}}]},
            >>>    {"skøde": [{"LOWER": "skøde"}]},
            >>> ]
            >>> match_counter.gen_matcher_objects_from_pattern_list(pattern_container_list)
```

dfm/description/gen_description.py
```diff
@@ -0,0 +1,247 @@
...
+
+        return aggregated_match_counts
+
+
+if __name__ == "__main__":
+    any_token_pattern = [{"tokens": [{"TEXT": {"REGEX": ".+"}}]}]
+
+    ############
+    # Religion #
+    ############
```
KennethEnevoldsen, line 161:
```suggestion
    # Religion
```

dfm/description/gen_description.py
```diff
@@ -0,0 +1,247 @@
...
+        {"muslim": [{"LOWER": {"REGEX": "muslim.*"}}]},
+        {"jew": [{"LOWER": {"REGEX": "jødi.*"}}]},
+        {"jew": [{"LOWER": {"REGEX": "(?!jødeskæg)jøde.*"}}]},  # Jødeskæg == Stueplante
+    ]
+
+    muslim_name_patterns = get_muslim_name_patterns()
+
+    ###########
+    # Genders #
+    ###########
```
KennethEnevoldsen, line 176:
```suggestion
    # Genders
```

dfm/description/gen_description.py
```diff
@@ -0,0 +1,247 @@
...
+    male_gendered_term_patterns = MatchCounter.term_list_to_lowercase_match_patterns(
+        male_gendered_terms, label="male_gendered_terms"
+    )
+    female_gendered_term_patterns = MatchCounter.term_list_to_lowercase_match_patterns(
+        female_gendered_terms, label="female_gendered_terms"
+    )
+
+    ###############
+    # Occupations #
+    ###############
```
KennethEnevoldsen, line 195:
```suggestion
    # Occupations 
```

dfm/description/gen_description.py
```diff
@@ -0,0 +1,247 @@
...
+    occupation_patterns = MatchCounter.term_list_to_lowercase_match_patterns(
+        occupation_pattern_list, label_prefix="occu_"
+    )
+    # List is a partial translation of Rae et al. 2022, p. 95
+
+    ###############
+    # Adult words #
+    ###############
```
KennethEnevoldsen, line 203:
```suggestion
    # Adult words
```

dfm/description/gen_description.py
```diff
@@ -0,0 +1,247 @@
...
+    adult_patterns = MatchCounter.term_list_to_lowercase_match_patterns(
+        danish_adult_words, label_prefix="porn_"
+    )
+
+    #############
+    # Execution #
+    #############
```
KennethEnevoldsen, line 210:
```suggestion
    # Execution
```

KennethEnevoldsen (changes requested):
I would add a test for the MatchCounter class which is currently completely untested. An rewrite the python file to a json.

requirements.txt
KennethEnevoldsen: Ahh. Jamen hold den på 0.8.2. Der er en PR på vej til DaCy - den er dog ikke som priority lige nu.

martbern (commented):
Misc. formatting

KennethEnevoldsen:
Mismatch in black version - update to the one in requirements in main

martbern:
Checks fail because of a dependency-incompatability with dacy. @KennethEnevoldsen, anything I can do here? 🙏

martbern:
```
ERROR: Cannot install -r requirements.txt (line 12), dacy, spacy<=3.3.0 and >=3.1.4 and transformers~=4.16.2 because these package versions have conflicting dependencies.
```

Huh, right now it's a triple-dependency problem. 

In `requirements.txt` we have:
```transformers~=4.16.2
spacy<=3.3.0 and >=3.1.4```

`dacy 1.3.0 depends on spacy<3.3.0 and >=3.2.0`
`spacy-transformers 1.0.6 depends on transformers<4.10.0 and >=3.4.0`

**Comment by martbern** [comment]
Yeah, that reduces the dependency conflict to:

`ERROR: Cannot install dacy and transformers~=4.16.0 because these package versions have conflicting dependencies.
[438](https://github.com/centre-for-humanities-computing/danish-foundation-models/runs/5534445530?check_suite_focus=true#step:4:438)

[439](https://github.com/centre-for-humanities-computing/danish-foundation-models/runs/5534445530?check_suite_focus=true#step:4:439)
The conflict is caused by:
[440](https://github.com/centre-for-humanities-computing/danish-foundation-models/runs/5534445530?check_suite_focus=true#step:4:440)
    The user requested transformers~=4.16.0
[441](https://github.com/centre-for-humanities-computing/danish-foundation-models/runs/5534445530?check_suite_focus=true#step:4:441)
    spacy-transformers 1.0.6 depends on transformers<4.10.0 and >=3.4.0`

**Comment by KennethEnevoldsen** [comment]
Thanks for looking into this. Yea so even changing to the newest version of spacy transformers would not solve the problem.

Something like [this](https://stackoverflow.com/questions/31798280/pass-no-deps-in-pip-requirements-txt) should though
```
# custom install of DaCy where we only use install the requirements needed for the lookup tables:
# should not be necessary after: https://github.com/centre-for-humanities-computing/DaCy/pull/98
dacy --install-option="--no-deps"
pandas>=1.0.0,<2.0.0
```
(in requirements.txt)

Then you can also remove the downgrade of wasabi.

File: dfm/description/matchcounter.py [hunk 17]
```diff
@@ -0,0 +1,108 @@
...
+        Get match counts for a list of SpaCy matcher-objects
+
+        args:
+            doc (Doc)
+            pattern_container_list (list): A list of dictionaries fitting SpaCy patterns
+            nlp: Language
+
+        returns:
+            A dictionary of the format {pattern_label (str): count (int)}.
+        """
```
KennethEnevoldsen, line 73: from the docstring:
```suggestion
    def _get_match_counts_from_doc(self, doc: Doc, matcher_object: Matcher) -> Dict[str, int]:
        """
        Get match counts for a list of SpaCy matcher-objects

        args:
            doc (Doc)
            pattern_container_list (list): A list of dictionaries fitting SpaCy patterns
            nlp: Language

        returns:
            Dict[str, int]: A dictionary of the format {pattern_label (str): count (int)}.
        """
```

.gitignore
```diff
@@ -131,8 +131,14 @@ dmypy.json
...
 .vscode/
 
-# JSONL data files
+# data files
 **/*.jsonl
+
+*.csv
+
+# MacOS
+.DS_Store
```
KennethEnevoldsen, line 140: duplicate?

.gitignore
```diff
@@ -131,8 +131,14 @@ dmypy.json
...
+=======
 *.ndjson
```
KennethEnevoldsen, line 142: move up to jsonl (essentially same format)

KennethEnevoldsen (changes requested):

dfm/description/gen_description.py
```diff
@@ -0,0 +1,135 @@
+import os
+import sys
+import time
+from datasets import load_dataset
+
+
+import spacy
+
+sys.path.append(".")
```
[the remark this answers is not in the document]
martbern: Ah, agreed! I've moved gen_description to `src/dfm` as it's probably more of an application script than general code. Moved the generalisable parts into `matchcounter` or `description_patterns`.

dfm/description/gen_description.py
```diff
@@ -0,0 +1,135 @@
...
+
+    ds = load_dataset("DDSC/partial-danish-gigaword-no-twitter")
+    ds_sharded = ds.shuffle()["train"].shard(
+        num_shards=100, index=0
+    )  # Work on 1/100th of DGW
+
+    nlp = spacy.blank("da")
+    nlp.max_length = 50000000
+
+    MatchCounter = MatchCounter(match_patterns=combined_patterns, nlp=nlp)
```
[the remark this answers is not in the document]
martbern: Hmm, sure, but the pattern selection is still application-specific, right? I've moved the generation into a function to make the code slightly more readable. Do we want to generalise it even further, and if so, why? :-)

martbern:
Thanks for the thorough code review! Dependencies are still failing, though:

`ERROR: dacy 1.3.0 has requirement wasabi<0.9.0,>=0.8.2, but you'll have wasabi 0.9.0 which is incompatible.` in spite of `--no-deps`?

`ERROR: spacy-transformers 1.0.6 has requirement transformers<4.10.0,>=3.4.0, but you'll have transformers 4.16.2 which is incompatible.` – which we don't have a solution for?

dfm/description/matchcounter.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    def __init__(self, match_patterns: List[Dict[str, list]], nlp: Language):
+        self.nlp = nlp
+        self.matcher_objects = self.gen_matcher_objects_from_pattern_list(
+            match_patterns
+        )
+
+    @staticmethod
+    def term_list_to_lowercase_match_patterns(
```
mollerhoj, line 25:
```suggestion
    def term_list_to_spacy_match_patterns(
```

Given the description of this method, I believe this is a more accurate name? Lowercasing seems like something one might wanna add as a flag to this method.

dfm/description/matchcounter.py
```diff
@@ -0,0 +1,129 @@
...
+            if label is None:
+                cur_label = label_prefix + term
+                out_list.append({cur_label: [{"LOWER": term}]})
+            else:
+                cur_label = label_prefix + label
+                out_list.append({cur_label: [{"LOWER": term}]})
+
+        return out_list
+
+    def gen_matcher_objects_from_pattern_list(
```
mollerhoj, line 55:
```suggestion
    def generates_matcher_objects_from_pattern_list(
```
Abbreviations are bad, non standard abbreviations are worse :-) https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764

dfm/description/matchcounter.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        out_list = []
+
+        for term in term_list:
+            if label is None:
+                cur_label = label_prefix + term
+                out_list.append({cur_label: [{"LOWER": term}]})
+            else:
+                cur_label = label_prefix + label
+                out_list.append({cur_label: [{"LOWER": term}]})
```
mollerhoj, line 51: Looks like there is an opportunity to reduce code duplication here - Can you find it?

dfm/description/matchcounter.py
```diff
@@ -0,0 +1,129 @@
...
+            >>>    {"atheism": [{"LOWER": {"REGEX": "atei.+"}}]},
+            >>>    {"skøde": [{"LOWER": "skøde"}]},
+            >>> ]
+            >>> match_counter.gen_matcher_objects_from_pattern_list(pattern_container_list)
+        """
+        matcher_object = Matcher(self.nlp.vocab)
+
+        for pattern_container in pattern_container_list:
+            pattern_label, subpattern_list = list(*pattern_container.items())
+
```
mollerhoj, line 74: This blank line strikes me as superfluous?

dfm/description/matchcounter.py
```diff
@@ -0,0 +1,129 @@
...
+        return matcher_object
+
+    def _get_match_counts_from_doc(self, doc: Doc, matcher_object: Matcher) -> dict:
+        """
+        Get match counts for a list of SpaCy matcher-objects
+
+        args:
+            doc (Doc)
+            pattern_container_list (list): A list of dictionaries fitting SpaCy patterns
+            nlp: Language
```
mollerhoj, line 86: arguments and description do not match?

dfm/description/matchcounter.py
```diff
@@ -0,0 +1,129 @@
...
+
+        return out_list
+
+    def gen_matcher_objects_from_pattern_list(
+        self, pattern_container_list: list
+    ) -> Matcher:
+        """
+        Generates matcher objects from a list of dictionaries with {matcher_label (str): pattern (list)}
+        Pattern must conform to SpaCy pattern standards:
+
```
mollerhoj, line 61: Should we not document the argument(s) to this method?

dfm/description/matchcounter.py
```diff
@@ -0,0 +1,129 @@
+from collections import defaultdict
+
+from spacy.language import Language
+from spacy.matcher import Matcher
+from spacy.tokens import Doc
+
+from typing import Dict, Iterable, List, Optional
+
+
+class MatchCounter:
```
mollerhoj, line 10: The name of this file should be `match_counter`, or the class should be `Matchcounter`. I prefer the former for obvious reasons.

tests/description/matchcounter_test.py
```diff
@@ -0,0 +1,51 @@
...
+        ]
+
+    def test_matcher_object_generation(self, regex_patterns, mc_basic):
+        matcher_objects = mc_basic.gen_matcher_objects_from_pattern_list(
+            pattern_container_list=regex_patterns
+        )
+
+        assert len(matcher_objects) == 1
+
+    def test_counting(self, texts, mc_basic):
```
mollerhoj, line 48: Could test be named/extended/documented to let readers know whats significant about the lists of integers? e.g. what does `[0, 1, 0]` mean?

tests/description/matchcounter_test.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+from dfm.description import MatchCounter
+import spacy
+
+
+class TestMatchCounter:
```
mollerhoj, line 6: A few suggestions for more descriptive/covering/documenting specs:

- Can we show what happens when a token occurs more than once?
- What happens with regarding to different cases? `Soldat`, `soldat`?
- What about normalization? `pæn` == `paen` == `pœn`?
- How do we handle symbols? `pæn,` == `pæn`? `torn-yester` == `tornyester`?
- How do we handle newlines? `pæn,` == `pæn`? `torn\nyester` == `tornyester`?
- Do we lemmatize? `krigen` == `krig`?

mollerhoj (changes requested):
Abbreviations should have been made illegal (and the punishment should be severe). Is "gen" short for gender? or gentrification? ;-)

KennethEnevoldsen:
I don't know how you do the local install? You should be using the `pip install -e .`

I don't know if you can specify `torch<1.11.0,>=1.10.2` try using `torch>=1.10.2,<1.11.0`, but there should be a working torch version for linux.

It seems like to no deps flag is not working in requirements.txt, but you should be able to resolve it in the setup.py which install the package. What you need to do here is remove the dependency from the requirements (and setup.cfg) and have it install dacy without dependencies (something like [this](https://stackoverflow.com/questions/12332975/installing-python-module-within-code)) in the setup.py.

dfm/description/matchcounter.py
KennethEnevoldsen: Obv. reason here being consistentency with the casing (indicating that it is two words).

src/dfm/description/gen_description.py
```diff
@@ -0,0 +1,114 @@
...
+    danish_adult_words,
+    get_muslim_name_patterns,
+    get_gender_name_patterns,
+)
+
+from dfm.description.matchcounter import MatchCounter
+
+
+def remove_irrelevant_columns(ds):
+    return ds.remove_columns(["text", "doc_id", "LICENSE", "uri", "date_built"])
```
KennethEnevoldsen, line 24: remove this function it saves you nothing

src/dfm/description/gen_description.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def create_patterns():
```
KennethEnevoldsen, line 27: Add docstring + return

src/dfm/description/gen_description.py
KennethEnevoldsen: actually, the line at the place and a comment would probably be more readable

martbern:
@KennethEnevoldsen, ready for mere? :-)

KennethEnevoldsen:
@martbern you can also press the "rerequest review" top right corner next to my name ;)

requirements.txt
```diff
@@ -1,8 +1,8 @@
 wandb~=0.12.9
 
 # models
-transformers~=4.16.2
-torch~=1.10.2
+transformers~=4.16.0
+torch>=1.6.0
```
KennethEnevoldsen, line 5: Add upper bound as we discussed, but it is good that the test pass.

requirements.txt
```diff
@@ -18,6 +18,10 @@ datasketch~=1.5.6
 google==3.0.0  # safesearch filtering
 more_itertools~=8.12.0
 
+# custom install of DaCy where we only use install the requirements needed for the lookup tables:
+# should not be necessary after: https://github.com/centre-for-humanities-computing/DaCy/pull/98
```
KennethEnevoldsen, line 22: Move comment to setup.py

KennethEnevoldsen (changes requested):
A few minor additions once they are fixed I feel it is ready, but @mollerhoj should probably also have a look.

src/dfm/description/gen_description.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Religion
+    religion_patterns = [
+        {"atheist": [{"LOWER": {"REGEX": "ath{0,1}eis.*"}}]},
+        {"buddhist": [{"LOWER": {"REGEX": "buddh{0,1}.*"}}]},
+        {"christian": [{"LOWER": {"REGEX": "kriste.*"}}]},
+        {"hindu": [{"LOWER": {"REGEX": "hindu.*"}}]},
+        {"muslim": [{"LOWER": {"REGEX": "muslim.*"}}]},
+        {"jew": [{"LOWER": {"REGEX": "jødi.*"}}]},
+        {"jew": [{"LOWER": {"REGEX": "(?!jødeskæg)jøde.*"}}]},  # Jødeskæg == Stueplante
```
mollerhoj, line 40: I always advice against using regexes when term lists will do. Regexes might be shorter to write but are more difficult to read and include you didn't consider. My dad last name is `kristensen`, but I can assure you he's no christian.

mollerhoj: I looks like we could transfor all data into term lists, and thus get rid of this custom function and make it easy to add new data (simply include a new list of terms in a data folder)

src/dfm/description/match_counter.py
```diff
@@ -0,0 +1,139 @@
...
+
+            out_list.append({cur_label: [{attribute: term}]})
+
+        return out_list
+
+    def create_matcher_objects_from_pattern_list(
+        self, pattern_container_list: List[Dict[str, List]]
+    ) -> Matcher:
+        """
+        Generates matcher objects from a list of dictionaries with {matcher_label (str): pattern (list)}
```
mollerhoj, line 64:
```suggestion
        Generates a matcher object from a list of dictionaries with {matcher_label (str): pattern (list)}
```

src/dfm/description/match_counter.py
```diff
@@ -0,0 +1,139 @@
...
+            if label is None:
+                cur_label = label_prefix + term
+            else:
+                cur_label = label_prefix + label
+
+            out_list.append({cur_label: [{attribute: term}]})
+
+        return out_list
+
+    def create_matcher_objects_from_pattern_list(
```
mollerhoj, line 60:
```suggestion
    def create_matcher_object_from_pattern_list(
```

src/dfm/description/match_counter.py
```diff
@@ -0,0 +1,139 @@
...
+        # Make sure that all elements are represented in the dict
+        for pattern in matcher_object._patterns:
+            pattern_label = self.nlp.vocab.strings[pattern]
+
+            counts[pattern_label] = 0
+
+        for match_id, start, end in matcher_object(doc):
+            counts[self.nlp.vocab.strings[match_id]] += 1
+
+        return dict(counts)
```
mollerhoj, line 113: Can we more this method down below the `count` method that uses it?

mollerhoj (changes requested):
I'm not sure why Spacy is needed here. I looks like we're simply counting word occurances without using any advanced processing?

If thats the case, then I think would could improve both code complexity and performance by simply doing a reverse index (keys as words, values as a list of categories the word belongs to). Scanning a doc and looking up words should run in practically O(n) as most words will belong to a small number of categories.

src/dfm/description/match_counter.py
KennethEnevoldsen: Not sure I understand?

martbern: I think @mollerhoj just wants to move `_get_match_counts_from_doc` below `count` due to their internal relationship 👍

martbern:
> I'm not sure why Spacy is needed for this. It looks like we're simply counting word occurrences without using any advanced processing?
> 
> If thats the case, then I think would could improve both code complexity and performance by simply using a reverse index (keys as words, values as a list of categories the word belongs to). Scanning a doc and looking up words should run in practically O(n) as most words will belong to a small number of categories.

Hmm, yeah, currently we aren't using any of the more advanced processing steps that Spacy affords us. 

That said, spacy also allows us to just use a term list – so maybe convert the regex to a term list, but maintain spacy in case we want to do something more complicated later? What do you guys think, @KennethEnevoldsen, @mollerhoj?

KennethEnevoldsen:
> I'm not sure why Spacy is needed for this. It looks like we're simply counting word occurrences without using any advanced processing?

You are completely correct. There is two main advantages here. 

1) SpaCy tokenizer is the best tokenizer for Danish text (at least tested on DDT) handling a wide variety of edge cases. Stanza's is also quite good, but as it is neural is fails in weird ways which is somtimes unpredictable. 

Why use match patterns instead of a dict then?

2) It allow us to quickly transition the extraction into using e.g. POS-tags. I allow us to do matches on sequence of text e.g. "___ was very happy". However I agree in its current form it seems silly, but the idea is that this approach extends easier simply forcing the user to write new match patterns as opposed to a list.

src/dfm/description/match_counter.py
mollerhoj: Yup, literally talking about how the line position of the methods in the file, nothing complicated really - just what I believe to be standard conventions.

mollerhoj:
Okay both those points make sense to me. I'm not sure how if the `O(n*m)` runtime of this approach is prohibitive or not, but it should be rather easy to parallelize if it becomes a problem.

martbern:
The huggingface `.apply ` has parallellisation across docs - I can do 10% of DAGW in 10 minutes on my M1 👍 

@KennethEnevoldsen, I think it's ready for merge – lemme know if you have corrections.

src/dfm/description/generate_description.py
```diff
@@ -0,0 +1,91 @@
+import os
+import sys
+import time
+from typing import List
+from datasets import load_dataset
+
+
+import spacy
+
+sys.path.append(".")
```
KennethEnevoldsen, line 10: Shouldn't be needed.

src/dfm/description/generate_description.py
```diff
@@ -0,0 +1,91 @@
...
+    dgw_processed = ds_sharded.map(
+        lambda batch: MatchCounter.count(batch["text"]),
+        batched=True,
+        batch_size=50,
+        num_proc=16,
+    )
+
+    print(f"\n\n--- Execution time was {time.time() - start_time} seconds ---")
+
+    from pathlib import Path
```
KennethEnevoldsen, line 81: move to start of if name ...

KennethEnevoldsen (approved):
A few minor changes. You can merge once they are fixed

mollerhoj:
> The huggingface `.apply` has parallellisation across docs - I can do 10% of DAGW in 10 minutes on my M1 👍
> 
> @KennethEnevoldsen, I think it's ready for merge – lemme know if you have corrections.

That sounds extremely slow unfortunately. Actually surprised that its that slow

KennethEnevoldsen:
On a single core or multiple cores?

martbern:
Huh, damn. That's with 16 jobs.

Some of the docs are _way_ larger than others, so it spends about 20% of the time waiting for the last thread.

Is it a dealbreaker for this?

KennethEnevoldsen:
Let us merge it for now and sample from the data it should be fast enough and 10% of DAGW in 10 minutes gives is a sample size of 1B words in <1 hour using 64 cpus - which should be more than enough to get an estimate of the balance in the corpora.

If performance becomes a problems we will go back and redo this.

This of course means that we can use this as a pornographic filter, but that was also never the intention with the implementation. If this shows a problem with pornographic domains, then we can implement a specific pornographic filter.

mollerhoj:
SGTM

martbern:
> SGTM

Acronyms are bad, non-standard acronyms are worse ;-)

Haha, thanks for the reviews!

PR #54, opened 2022-02-09
Data description of QualityFilter

KennethEnevoldsen:
Data description of QualityFilter. 

This is a work in progress.

@mollerhoj I integrated the updated test, will you take a look at them?


mollerhoj (approved):
Looks fine. Though in the future I'd recommend not incorporating other PRs, it makes reviews a little harder IMO. What's the plan moving forward with this draft? What's missing? (Besides fixing the tests)?

src/examine_quality_filter.py/Repetitious_text.py
```diff
@@ -0,0 +1,115 @@
+"""
```
mollerhoj, line 1: Any reason for the new `src` folder? :D

KennethEnevoldsen: I actually went back and forth on this.

Basically, we kinda want two different code bases one which is simply the application, acting mostly as reference and documentation and the package which is intended to be reuseable.

KennethEnevoldsen: I would love if you have an opinion on this, in other forks I have actually done it using src/dfm and src/application

PR #61, opened 2022-02-24
Feat/deduper additions

saattrupdan:
This PR introduces several improvements and changes to the deduplication pipeline.

The major changes are the following:

1. Implemented lossless parallel deduplication. My speed tests showed that the slowest part of the deduplication is the computation of the MinHash fingerprints, which can be easily parallelised. This is lossless as we're working with the same fingerprints, and we're still deduplicating one document at a time (which is quite fast anyway).
2. When building the MinHash fingerprint for a document, I now use `update_batch` from the `datasketch` package, rather than calling `update` in a for-loop. This is a recent addition to `datasketch` and uses vectorisation, speeding things up by ~4x.

It also includes the following smaller changes:

1. Separate out `_get_shingles` method for easier readability.
2. Removed `char_ngram`, as it is simply way too slow and requires special needs to avoid memory errors, which I didn't think was worth it.
3. To enable to multiprocessing I am batching the dataset. To make sure that this still works if the dataset is iterable, I'm using the `chunked` function from the `more_itertools` package, which acts on iterables rather than, say, lists. Included this in requirements.


dfm/cleaning/deduper.py
```diff
@@ -64,50 +76,59 @@ class Deduper:
 
     def __init__(
         self,
-        split_method: Optional[str] = "char_ngram",
+        split_method: Optional[str] = "word_ngram",
```
mollerhoj, line 79: Ændring af default her kommer nok til at ødelægge tests?

dfm/cleaning/deduper.py
```diff
@@ -116,14 +137,35 @@ def _get_minhash(self, doc: str) -> LeanMinHash:
             ]
         elif self.split_method == "paragraph":
             shingles = [p for p in doc.split("\n") if len(p) > 0]
-        elif self.split_method == "none":
+        elif self.split_method == "none" or self.split_method is None:
```
mollerhoj, line 140: We might wanna remove the option to give a "none" string?

mollerhoj (approved):

mollerhoj:
A little sad to lose char_ngram - I think it could be great for other domains (twitter data, chatbots)

dfm/cleaning/deduper.py
saattrupdan: It's quite convenient though, and solves confusion. Say, if I want to create a simple CLI, then input will always be a string, and that enables me to simply write 'none'.

dfm/cleaning/deduper.py
saattrupdan: Yep, planning to change the tests to fit this change 👍

saattrupdan:
> A little sad to lose char_ngram - I think it could be great for other domains (twitter data, chatbots) - I realize it's not important for this specific use case

@mollerhoj Sure, but if one really wanted it for another project (or any other shingle-method for that matter), then it's quite simple to implement. I'd vote for including the things that we care about in this project, to avoid a cluttered codebase as much as possible.

mollerhoj:
> > A little sad to lose char_ngram - I think it could be great for other domains (twitter data, chatbots) - I realize it's not important for this specific use case
> 
> @mollerhoj Sure, but if one really wanted it for another project (or any other shingle-method for that matter), then it's quite simple to implement. I'd vote for including the things that we care about in this project, to avoid a cluttered codebase as much as possible.

Fair.

dfm/cleaning/deduper.py
KennethEnevoldsen: Hmm, I would probably argue for removing the NoneType as an option, which would make sure that the split_method is always a string.

KennethEnevoldsen (approved):

KennethEnevoldsen:
As always very clean code. I made a minor comment on removing the NoneType option for the string_method, but it is very minor.

PR #62, opened 2022-02-27
Fix deduper

saattrupdan:
Fixes a few things with the deduper:

1. The corpus should be converted to an iterable, not a list - otherwise the entire corpus would be loaded into memory!
2. There were methods with duplicate names - removed the old versions

Also modified docstrings slightly.


dfm/cleaning/deduper.py
```diff
@@ -240,7 +243,7 @@ def deduplicate(
 
         # Convert corpus to an iterable of strings if a Dataset is given
         if isinstance(corpus, Dataset) or isinstance(corpus, IterableDataset):
-            corpus = [sample["text"] for sample in corpus]
+            corpus = (sample["text"] for sample in corpus)
```
KennethEnevoldsen, line 246: I wondered why this changed originally, should have asked then. I will make sure to do so next time

KennethEnevoldsen (approved):

PR #71, opened 2022-03-07
Deduper checkpoints

saattrupdan:
Enable checkpointing of the deduper. This saves the internal state of the deduper during deduplication, so that it can be reloaded in case the script crashes.

Usage:
```python
>>> deduper = Deduper.load_from_disk('path/to/deduper/dir')
>>> deduper.deduplicate(corpus)
```

Further, the internal LSH cache is now stored as `deduper.lsh_cache`, a list of records `dict(id: int, duplicate: bool)` is stored in `deduper.mask` and the configuration can be fetched with `deduper.get_config()`. During deduplication these are stored in the `output_dir` folder, which is specified during initialisation.

If one does not want to reset these internal variables then simply call `deduper.reset()`. This does not affect the files on disk, however.

### DIFF
--- a/dfm/cleaning/deduper.py
+++ b/dfm/cleaning/deduper.py
@@ -17,6 +17,7 @@ from datasets.arrow_dataset import Dataset
 from datasets.iterable_dataset import IterableDataset
 from typing import Union, Iterable, Optional, List, Callable
 from pathlib import Path
+import shutil
 import json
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen (approved):
Looks good. Thanks for this update!

HLasse (approved):
Nice update, great functionality to have!

PR #76, opened 2022-03-09
update workflow with new version of covcomment

KennethEnevoldsen:
Update workflow with new version of covcomment, due to the previous version failing.


saattrupdan (approved):
LGTM!

PR #81, opened 2022-03-29
fix: Deal with `num_docs` when corpus is an iterable

saattrupdan:
This was an issue when the corpus was an iterable, as `num_docs` would be `None` in that case.


KennethEnevoldsen:
Looks good. Similar to the fix I made.

PR #83, opened 2022-03-29
Fix/dedup mask saving

saattrupdan:
This saves the current configuration, mask and LSH cache to disk when `overwrite` is True. This means that when an existing deduper is loaded with `load_from_disk` then it's possible to continue appending to the previous file.


KennethEnevoldsen (approved):
Looks good, thanks for the quick fix.

Issue #90, opened 2022-04-13
Fix OOM error for deduplication.


mollerhoj:
The deduplicator currently run out of memory when running on huge datasets.

***PROPOSAL*** (Other solutions are more than welcome)

Here's a (naive brute force) algorithm that sacrifices speed for memory:

- Run the deduplication on the data until RAM is (near) full. Mark the current position in the data.
- Continue to run the algorithm on the remaining data, but do not add hashes to the store.
- Clear the cache (free all memory), and jump to the previously marked position.
- Repeat.

It will run in polymorphic time O(n^2)/2, where n is the number of splits of the data required.

It might help to imagine running this on a batch size of 1. Then it's simply a O(n^2)/2 algorithm where n is the number of hashes (which is the naive bruteforce solution to the dedup problem and obviously not viable)

saattrupdan:
@mollerhoj How would you detect when memory is near full, as we want to automate this?

mollerhoj:
Oh I'd just set a max number of keys in the cache. As the hashes are all the same size we could make a parameter `lsh_cache_max_byte_size` and then calculate the `lsh_cache_max_keys = lsh_cache_max_byte_size / hash_byte_size` for ease of usage.

Usually you'd expect the user to know how much RAM we're allowed to grab...

KennethEnevoldsen:
Closing this issue as the PR was deemed stale.

PR #98, opened 2022-04-15
Deduper memory algo

mollerhoj:
An example of the algorithm described here: https://github.com/centre-for-humanities-computing/danish-foundation-models/issues/90

TODO:
- Do not store corpus if using naive algo (cache_limit_set)
- Cleanup: Cleanup tests, more tests, remove prints.

### DIFF
--- a/src/dfm/cleaning/deduper.py
+++ b/src/dfm/cleaning/deduper.py
@@ -466,8 +466,6 @@ class Deduper:
             with config_path.open("wb") as f:
                 pickle.dump(config, f)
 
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
Hei @mollerhoj should we polish this up and add it. Seems like a good addition, which we could def. use if we want to do a joint deduplication over the whole corpus.

KennethEnevoldsen:
Hej @mollerhoj seems like this PR has gone stale, is it ok if I close it?

mollerhoj:
sure

PR #99, opened 2022-04-17
patch memory issue caused by tee

KennethEnevoldsen:
Patched memory issue caused by the tee operator.

The deduplication script seems to have a large memory overhead, which isn't halved (as would have been expected) by reducing the `num_min_hash` from 128 -> 64. From this post [here](https://stackoverflow.com/questions/69754473/implicit-memory-consumption-with-itertools-tee-of-generators). It seems like the tee operator has to be emptied otherwise the duplicate is kept in memory.

This solution removed the `tee` and extract the first element for the corpus iterator to normalize the iterator, then add it back if using `itertools.chain`

### DIFF
--- a/src/dfm/cleaning/deduper.py
+++ b/src/dfm/cleaning/deduper.py
@@ -410,6 +410,7 @@ class Deduper:
         else:
 
             # extract the first element of the corpus
+            corpus = iter(corpus)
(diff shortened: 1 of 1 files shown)


saattrupdan (approved):
LGTM!

PR #100, opened 2022-04-19
Change import to enable running script as a CLI

KennethEnevoldsen:
Change import to enable running script as a CLI. E.g. 

```
python src/dfm/cleaning/dedupe.py -s paragraph --n_jobs 16
```

### DIFF
--- a/src/dfm/cleaning/deduper.py
+++ b/src/dfm/cleaning/deduper.py
@@ -27,7 +27,7 @@ from joblib import Parallel, delayed
 import multiprocessing as mp
 import pickle
 from functools import partial
-from .deduper_utils import get_minhash, default_normalization
+from dfm.cleaning.deduper_utils import get_minhash, default_normalization
 
 
 class Deduper:


saattrupdan:
@KennethEnevoldsen This will break imports though? Can't you just run the CLI as `python -m src.dfm.cleaning/dedupe -s paragraph --n_jobs 16`?

KennethEnevoldsen:
No, the import is fine with this change (python knows it is a part of a package). Didn't know the other approach, does that work?

saattrupdan:
> No, the import is fine with this change (python knows it is a part of a package). Didn't know the other approach, does that work?

Yep, I just tried it on my machine, and it works just fine :)

Generally we have to decide whether `dfm` is a package (in which case all imports should be relative) or a script (in which case all imports should be absolute). Since we've set everything up as a package, I vote for relative imports everywhere within the `dfm` folder.

KennethEnevoldsen:
With the `python -m src.dfm.cleaning.dedupe -s paragraph --n_jobs 16` solving the problem, I agree that it is better to keep it as is.

PR #101, opened 2022-04-20
Add test set tasks

saattrupdan:
Created plan for the test dataset.

Chip in with comments and/or additions to the list of test set tasks 🙂

### DIFF
--- a/test-dataset-plan.md
+++ b/test-dataset-plan.md
@@ -1,0 +1,39 @@
+# Plan for the test dataset
+
+We want the test dataset to contain an equal amount of data split into various
+tasks, described below.
+
+Aside from ensuring that the resulting models perform well across multiple
+tasks and do not overfit to any single one such, this would also provide
+transparent performance descriptions of the trained models. Such transparency
+would mean that anyone who is looking for a language model to finetune on their
+dataset can immediately see whether the model works well on their domain/task.
+
+## Tasks
+
+As mentioned above, we want the test dataset to contain an equal amount of
+data from each of the following tasks:
+
+- Danish web data (Netarkivet)
+- Danish social media data (HOPE Twitter, Reddit-da and Hestenettet from Danish
+  Gigaword)
+- Danish legal data (Retsinformation and Skat.dk, both from Danish Gigaword)
+- Danish wiki data (Danish Wikipedia and lex.dk)
+- Danish news data (DaNews)
+- Danish books (Gutenberg, Danish literature, WikiBooks and WikiSource, all
+  from Danish Gigaword)
+- Danish spontaneous text (OpenSubtitles and Spontaneous Speech, both from
+  Danish Gigaword)
+- Danish dialectal data (Botxt and Sønderjysk, both from Danish Gigaword)
+
+We would also like to test separately on other languages, to see how well it
+has understood these. These only appear as separate test scores and are not
+included in any aggregated test score of the model. To keep these scores
+comparable, they have all been taken from the same
+[OSCAR](https://ids-pub.bsz-bw.de/frontdoor/index/index/docId/9021) dataset.
+
+- English data ([OSCAR-en](https://huggingface.co/datasets/oscar/viewer/unshuffled_deduplicated_en/train)
+- Norwegian Bokmål data ([OSCAR-nb](https://huggingface.co/datasets/oscar/viewer/unshuffled_deduplicated_no/train))
+- Norwegian Nynorsk data ([OSCAR-nn](https://huggingface.co/datasets/oscar/viewer/unshuffled_deduplicated_nn/train))
+- Swedish data ([OSCAR-sv](https://huggingface.co/datasets/oscar/viewer/unshuffled_deduplicated_sv/train))
+- Icelandic data ([OSCAR-is](https://huggingface.co/datasets/oscar/viewer/unshuffled_deduplicated_is/train))
--- a/test-dataset-plan.md
+++ b/test-dataset-plan.md
@@ -9,7 +9,7 @@ transparent performance descriptions of the trained models. Such transparency
 would mean that anyone who is looking for a language model to finetune on their
 dataset can immediately see whether the model works well on their domain/task.
 
-## Tasks
+## Tasks
 
 As mentioned above, we want the test dataset to contain an equal amount of
 data from each of the following tasks:
--- a/test-dataset-plan.md
+++ b/test-dataset-plan.md
@@ -14,17 +14,18 @@ dataset can immediately see whether the model works well on their domain/task.
 As mentioned above, we want the test dataset to contain an equal amount of
 data from each of the following tasks:
 
-- Danish web data (Netarkivet)
-- Danish social media data (HOPE Twitter, Reddit-da and Hestenettet from Danish
-  Gigaword)
+- Danish web data (OSCAR-da)
+- Danish social media data (HOPE Twitter and Reddit-da)
 - Danish legal data (Retsinformation and Skat.dk, both from Danish Gigaword)
+- Danish clinical data (PSYCOP project)
 - Danish wiki data (Danish Wikipedia and lex.dk)
 - Danish news data (DaNews)
 - Danish books (Gutenberg, Danish literature, WikiBooks and WikiSource, all
   from Danish Gigaword)
 - Danish spontaneous text (OpenSubtitles and Spontaneous Speech, both from
   Danish Gigaword)
-- Danish dialectal data (Botxt and Sønderjysk, both from Danish Gigaword)
+- Danish dialectal data (Botxt, from Danish Gigaword)
+- Danish speeches (FTSpeech)
 
 We would also like to test separately on other languages, to see how well it
 has understood these. These only appear as separate test scores and are not


test-dataset-plan.md
```diff
@@ -0,0 +1,39 @@
...
+would mean that anyone who is looking for a language model to finetune on their
+dataset can immediately see whether the model works well on their domain/task.
+
+## Tasks
+
+As mentioned above, we want the test dataset to contain an equal amount of
+data from each of the following tasks:
+
+- Danish web data (Netarkivet)
+- Danish social media data (HOPE Twitter, Reddit-da and Hestenettet from Danish
```
martbern, line 18: I imagine Twitter data is quite different from Hestenettet, and that some applications would look into Twitter without ever touching something like Hestenettet. 

Perhaps it'd be helpful to report results for Twitter/Hestenettet (etc., whatever our lowest level of granularity is) separately?

martbern (commented):
Awesome Dan, looking good!

test-dataset-plan.md
```diff
@@ -0,0 +1,39 @@
...
+  Gigaword)
+- Danish legal data (Retsinformation and Skat.dk, both from Danish Gigaword)
+- Danish wiki data (Danish Wikipedia and lex.dk)
+- Danish news data (DaNews)
+- Danish books (Gutenberg, Danish literature, WikiBooks and WikiSource, all
+  from Danish Gigaword)
+- Danish spontaneous text (OpenSubtitles and Spontaneous Speech, both from
```
KennethEnevoldsen, line 25: Is relevant to include FT speech because it is typically used for speech to text?

KennethEnevoldsen (commented):
Looks good. I added a clinical domain text. I agree with Martin on the Twitter, but otherwise don't have any strong opinions.

How much do we want of each/in total? With the current groups an even split seems fine. Potentially a split within dialect such that each dialect accounts for the same (but dialect as a whole is the same as "clinical?") 

We also seems to be missing historic text (implicit in Gutenberg and Danish literature).

test-dataset-plan.md
KennethEnevoldsen: FT = folketinget

test-dataset-plan.md
```diff
@@ -0,0 +1,39 @@
...
+  Danish Gigaword)
+- Danish dialectal data (Botxt and Sønderjysk, both from Danish Gigaword)
+
```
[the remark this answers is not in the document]
saattrupdan: Again, we can add tonnes of different datasets - why clinical? In that case we could add all sorts of specialised expert data. I know you guys are into your health data, but I'm not sure if it makes sense to include it in a test set. If we _do_ want to include such data then we could bulk it with expert data from other fields (law, finance, Danish research articles etc., but since we don't have this atm, I'd leave it for future work).

test-dataset-plan.md
saattrupdan: Speeches could potentially make sense, but I don't want it to be included in the spontaneous text, since it's the opposite. The reason why spontaneous text could be interesting is because of all the discontinuities ("Hej, øh, jeg hedder Dan, hvad hed-, jeg mener, øh, hvem er du?")

test-dataset-plan.md
saattrupdan: In that case I'd vote for just removing Hestenettet, as I don't see the merit of it on its own.

saattrupdan:
I suppose a separate, but related, discussion is whether the test set should be public. This would make sense, I think, as it allows comparison of future models with this one. Most of the tasks are currently public, with the exception of news data and web data. The web data could be fixed by instead using part of mC4-da or OSCAR-da, say. I'm not sure what could substitute the news data - are there are open Danish news datasets out there?

HLasse:
> EDIT: I suppose the Twitter data is not exactly open either, so we could consider just using Reddit-da as substitute for that. Alternatively, we could anonymise the Twitter data and release that? I.e., remove all links, @-mentions and metadata.

I don't think we can share that either, however, we should be able to share tweet ids so others can rehydrate the data. Given the relatively large amount of Twitter data in the corpus I think we should at least keep part of it for testing. Potentially just some of what's included in DAGW (would need to rehydrate the ids).

test-dataset-plan.md
HLasse: +1 for separating Twitter. Fine with removing Hestenettet

test-dataset-plan.md
HLasse: Speeches could be interesting - agree on placing them in another group than spontaneous. FT Speech has some weird pecularities so it might be better to use speeches from e.g. the prime minister (I scraped them at some point so have them lying around).

HLasse (commented):
LGTM

test-dataset-plan.md
KennethEnevoldsen: I agree as well. @saattrupdan it should be possible to create the dataset downloader such that it rehydrates the tweets when downloading it (at least from what I can see in the documentation you can use the `use_auth_token` (or potentially even the **kwargs) to supply an optional Twitter token.

test-dataset-plan.md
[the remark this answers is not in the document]
KennethEnevoldsen: I think two of the main application areas in industry at the moment are ClinicalNLP and LegalNLP. I think it is fair to build specifically into this. Especially as I think the test set should be actionable, that is that a practitioner should be able to decide whether it is worth using the model for their task given the performance in their domain. That being said @HLasse and I plan to do a domain-specific fine-tuning of a DFM model for The Psych. domain so it can def. be left for that paper.

test-dataset-plan.md
KennethEnevoldsen: So keep spont. speech and speeches?

KennethEnevoldsen:
**Twitter:** Seems like HF datasets allow passing auth. tokens, which would allow one to rehydrate the test set on the fly. The problem is naturally that it makes a varying test set (as stuff gets deleted).

**Open vs closed:** I think it the relevant to have both an open test set, but I also think it is relevant to e.g. test the performance on news articles (as the opposite seems odd), however, it might be possible to do that simply from using DaNewsRoom and oscar? I think I prefer a more precise/varied test set than an exclusively open test set. Naturally, the open part can always be shared.

mollerhoj (approved):

test-dataset-plan.md
```diff
@@ -0,0 +1,39 @@
...
+- Danish social media data (HOPE Twitter, Reddit-da and Hestenettet from Danish
+  Gigaword)
+- Danish legal data (Retsinformation and Skat.dk, both from Danish Gigaword)
+- Danish wiki data (Danish Wikipedia and lex.dk)
+- Danish news data (DaNews)
+- Danish books (Gutenberg, Danish literature, WikiBooks and WikiSource, all
+  from Danish Gigaword)
+- Danish spontaneous text (OpenSubtitles and Spontaneous Speech, both from
+  Danish Gigaword)
+- Danish dialectal data (Botxt and Sønderjysk, both from Danish Gigaword)
```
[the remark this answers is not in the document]
MalteHB: I think we should seperate them. They are VERY different in linguistic representation. Furthermore, according to the discussion from wednesday meeting, I would really like to see what difference including Botxt and other dialects, that seem to be a completely different language, in the model training.

test-dataset-plan.md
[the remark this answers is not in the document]
saattrupdan: Ah right, fair enough. But sure, having two expert-style texts could make sense, I was just worried that the same argument could be made for every specialised industry, but I get your point in including the two most popular ones. I'll update that.

test-dataset-plan.md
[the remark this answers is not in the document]
saattrupdan: @MalteHB Alright. In that case I vote to just remove the `Sønderjysk` dataset, as its too small to stand on its own (it's only 20kb, where `Botxt` is 400kb).

test-dataset-plan.md
saattrupdan: @MalteHB There is 700kb spontaneous speech in DAGW, and you get a bunch from web data, Twitter data, reddit and hestenettet, too. I'm not worried about a lack of training data, just that we choose the right test data :)

saattrupdan:
Changed the following now:
- Changed Netarkivet to OSCAR-da
- Removed Hestenettet and Sønderjysk
- Added clinical data and FTSpeech

Does this sound good with everyone?

@KennethEnevoldsen @MalteHB @HLasse

MalteHB:
What about Bornholmsk?

saattrupdan:
> What about Bornholmsk?

@MalteHB Still in there (Botxt dataset).

martbern (approved):

KennethEnevoldsen (commented):
Why was Sønderjysk removed? EDIT: Ahh I see since it was too small
In that case, I think I would prefer it as a test set.

Otherwise looks good. The size is still missing, but that is another PR

Edit: Wikitext3 uses 100 million tokens (~10% of DAGW). Given 10 categories that is 10k tokens pr. category.

KennethEnevoldsen (approved):

PR #178, opened 2023-10-12
ci: update repo structure and CI/CD for v2

MartinBernstorff:
# General considerations
I've made a bunch of decisions below, and some of them are based on taste.

By far the most important criteria are:
* High developer enjoyment
* High productivity

So definitely raise any concerns you have!

I also like to geek out on this, so would love to have a talk about it ;-)

### Static type checking
Pyright. Primary advantage over mypy is that mypy infers any type that is not hinted as "Any", whereas Pyright defines its own "Unknown" type. For more, see https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md

I've set it to the "strict" setting, which is (unsurprisingly) quite strict. This meant the existing code had ~400 errors.

"Strict" matches my preference, but it is definitely a strong requirement for developers. I'm super open to opinions here.
### Archiving
Instead of trying to fix all of the errors, I spoke with Lasse Hansen, and we agreed to move it to an "archive" folder, and then we can extract functionality into the src folder as we need it.
### Dev env
Since we're writing an application, rather than a library with many users, we can take a bunch of liberties that makes maintenance easier. For example:
* Only one Python version (3.10 for now)
* Pin all dependencies to one version
* Specify the entire development environment in a dev container. This ensures there is only one source of truth.

I've specified how to spin up the dev environment in the readme. This also installs relevant VSCode extensions.

It also allows us to use the dev container for CI, ensuring complete alignment between local and CI environments. It currently takes 1,5 minutes to build, but that'll be cached when we push to main the first time.

Classical installation is still possible; all you have to set up are the requirements-files and the package itself.

### Formatting
Ruff and black. Nothing wild here.

### Tests
Pytest. Again, nothing wild. I'll argue for including codecov as well; not because we should follow it religiously, but because it often flags code that I forgot to test sufficiently.

### Dependency management
KISS: requirements.txt files.

Split out into a few to get more efficient caching when building the docker image.

@KennethEnevoldsen, @saattrupdan, could you ping Rasmus and Peter as well?

### Changes needed:
- [x] Add an `Install` entrypoint
- [ ] Add a "open in dev container" button
- [x] Migrate dependencies to pyproject.toml


saattrupdan (approved):
Some small things, otherwise looks fine.

Also, I cannot add Rasmus and Peter as reviewers since I'm not part of chcaa. Are you able to make us admins?

I'm going on a holiday for a week today, so can't answer any follow-ups until I'm back.

archive_v1/requirements.txt
```diff
@@ -0,0 +1,31 @@
...
+# utility
+wasabi>=0.9.0,<0.10.0
+pydantic~=1.8.2
+pre-commit~=2.17.0
+
+
+
+# format
+black>=22.3.0,<22.4.0
+isort~=5.10.1
```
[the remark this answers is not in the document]
KennethEnevoldsen: Ruff [does not replace black](https://docs.astral.sh/ruff/faq/#is-ruff-compatible-with-black). I would install it here as well as you want your ide formatter to use the same version (and it will default to use the black in the environment).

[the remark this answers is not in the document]
MartinBernstorff: Agreed with Kenneth! Also, running them natively ensures:
* Alignment between IDE-formatter, CLI and CI
* Faster performance, since it doesn't need to activate a .venv (pre-commit installs each utility in its own .venv)

archive_v1/requirements.txt
```diff
@@ -0,0 +1,31 @@
+wandb~=0.12.9
+
+# models
+transformers>=4.20.1,<4.21.0
+torch>=1.6.0,<1.13.0
```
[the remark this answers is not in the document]
MartinBernstorff: Just used the version from the previous requirements file. Fine by me to pin the version to the newest version. @KennethEnevoldsen, I assume you agree? (I assume we won't need to run this on Ovartaci).

MartinBernstorff:
@KennethEnevoldsen Are we up for trying out codecov? :-)

archive_v1/requirements.txt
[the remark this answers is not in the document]
KennethEnevoldsen: it is an archived req.

requirements.txt
KennethEnevoldsen, line 1: @MartinBernstorff why not have the requirements in the pyproject.toml?

archive_v1/requirements.txt
KennethEnevoldsen, line 1: I would move this either to an independent branch, integrate it or remove it (I don't want to keep it as a folder)

KennethEnevoldsen (commented):
I think it looks good, but I would add more documentation on how to e.g. run the make commands. 

@saattrupdan does everyone use vs-code?

requirements.txt
MartinBernstorff: No strong opinion here. Primary concern is lack of cross-dependency-manager support (e.g. poor support for anaconda, and different syntax depending on poetry or pip), but that is definitely a minor concern.

I'm fine with adding it to the pyproject.toml.

KennethEnevoldsen:
@MartinBernstorff: Sure thing we could try out codecov

archive_v1/requirements.txt
MartinBernstorff: Why though?

Advantage of the current solution is similar to the advantage of having a monorepo; it's super easy to migrate something from the archive to "production". I think we should keep it as a folder, at least to begin with, and then we can delete it and use a historical commit ID as the archive.

archive_v1/requirements.txt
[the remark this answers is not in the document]
MartinBernstorff: Not sure which question you're answering here? Pin to newest version or delete?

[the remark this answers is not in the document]
MartinBernstorff: Ah, I see, these are requirements for the archived code! We should add the newest version of torch as a pinned dependency when we need it.

saattrupdan:
> I think it looks good, but I would add more documentation on how to e.g. run the make commands. 
> 
> 
> 
> @saattrupdan does everyone use vs-code?

I use nvim, so nopes.

requirements.txt
saattrupdan: I agree that pyproject is a bit nicer. I haven't used it with pip - how does that work? Can you set it up so that it appears in pyproject when you pip install?

archive_v1/requirements.txt
KennethEnevoldsen: Ahh right, I would agree with that approach

requirements.txt
KennethEnevoldsen: I would prefer .toml, as it allows me to package the repo more easily. The preference is minor though.

KennethEnevoldsen: pyproject.toml is fully pip installable (unless you are using older versions of pip). The formatting is slightly different compared to poetry, but generally, I would say it is very similar.

KennethEnevoldsen: I would prefer .toml, but. it is only a slight preference

archive_v1/requirements.txt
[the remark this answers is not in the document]
KennethEnevoldsen: Exactly ^

KennethEnevoldsen:
It might be worth updating the python version

MartinBernstorff:
Devcontainers are supported by nvim as well: https://github.com/esensar/nvim-dev-container

Or through a CLI: https://code.visualstudio.com/docs/devcontainers/devcontainer-cli

KennethEnevoldsen:
@MartinBernstorff I just want to make sure that the documentation of the setup is clear regardless of the IDE.

requirements.txt
MartinBernstorff: Using a .toml file means that all dependencies are cached in the same layer when building the Docker image. Whenever we change a dependency, the whole layer needs to be rebuilt (1-2 minutes depending on dependencies).

This is only a concern for those using the Docker image, so a trade-off. I'll see what the others prefer 👍

MartinBernstorff:
> @MartinBernstorff I just want to make sure that the documentation of the setup is clear regardless of the IDE.

Agreed. We can update based on what the others say re: requirements.txt vs. pyproject.toml. Ideally, we should have only one entrypoint for a full dev env setup, e.g. dev container that uses `make install` or similar.

requirements.txt
KennethEnevoldsen: So it is an argument for coupling/decoupling docker and built for a benefit of 1-2 minutes when you change the dependencies. Then I would go for decoupling.

rlrs:
One perspective from my side, is that most things we make here will have to run on some sort of HPC infrastructure. In order to have access to a decent software stack, we will have to use either a container or the module system. On LUMI it's definitely recommended to use a container. The one I have currently based everything on is https://hub.docker.com/r/rocm/pytorch/ with the tag `rocm5.4.2_ubuntu20.04_py3.8_pytorch_2.0.0_preview`. 

Since all the packages we're going to compile/install will depend on the software stack in the container, we should probably make this explicit in this repo. It for example means that the Python version is going to depend on the container, so it should be tied to that rather than some other place in the repo?

KennethEnevoldsen:
@rlrs hvorfor ikke [rocm5.7_ubuntu22.04_py3.10_pytorch_2.0.1](https://hub.docker.com/layers/rocm/pytorch/rocm5.7_ubuntu22.04_py3.10_pytorch_2.0.1/images/sha256-21df283b1712f3d73884b9bc4733919374344ceacb694e8fbc2c50bdd3e767ee?context=explore)?

requirements.txt
MartinBernstorff: I don't think it's a coupling argument. My understanding is that if two elements, X and Y, are more tightly coupled, it implies that a change in either is more likely to require a change in the other.

Using requirements.txt over pyproject.toml actually decouples our dependency specification. It means we can change our dependency resolves between pip, conda or poetry with fewer changes than if we used pyproject.toml.

That said, the prior arguments stil hold, so I'm open to using .toml.

MartinBernstorff:
Great points @rlrs! Just a few clarifying questions:

> we will have to use either a container or the module system

Unsure what you're pointing towards when saying "module system" here. Also, is this an either/or?

> It for example means that the Python version is going to depend on the container, so it should be tied to that rather than some other place in the repo?

Completely agree. One source of truth ftw. If the others agree, I'll remove as many of the python versions specified in the pyproject.toml as possible.

> With that said, my current approach is to create a virtualenv on the host that then gets activated in the container, so you can upgrade packages without rebuilding the container

You can also connect to the container and update the dependencies within it. This removes the need for a virtualenv, and ensures that the deps are on the same filesystem. Especially on macOS/Windows, it should give quite a performance boost compared to bind-mounts :-)

> quite a few packages nonetheless have to be compiled against a specific ROCm stack

Another argument for installing deps within the container.

If we end up needing a more complicated container stack, e.g. one for ROCm and one for CUDA, this setup should support it. But let's cross that bridge when we get to it?

rlrs:
> @rlrs hvorfor ikke [rocm5.7_ubuntu22.04_py3.10_pytorch_2.0.1](https://hub.docker.com/layers/rocm/pytorch/rocm5.7_ubuntu22.04_py3.10_pytorch_2.0.1/images/sha256-21df283b1712f3d73884b9bc4733919374344ceacb694e8fbc2c50bdd3e767ee?context=explore)?

Didn't exist when I started, and I had trouble with the newest one at that point (with 5.6). Can look into it again at some point, but I mostly used 5.4.2 because I knew people had that working. Not sure it makes much difference for now, anyways.

> Unsure what you're pointing towards when saying "module system" here. Also, is this an either/or?
LUMI has a [module system](https://docs.lumi-supercomputer.eu/runjobs/lumi_env/Lmod_modules/) which allows you to build/load custom software packages. But I definitely think containers are the nicer way to do it.

> You can also connect to the container and update the dependencies within it. This removes the need for a virtualenv, and ensures that the deps are on the same filesystem. Especially on macOS/Windows, it should give quite a performance boost compared to bind-mounts :-)

I don't think Singularity (the container system on LUMI) is quite that flexible. It's not like Docker and you even need root for some of the features, which we cannot have. But maybe it's possible and I just don't know how.

peterbjorgensen (approved):
Looks fine. I am not sure containers run well on multi-node-slurm systems, but @rlrs will now better.

MartinBernstorff:
> I don't think Singularity (the container system on LUMI) is quite that flexible. It's not like Docker and you even need root for some of the features, which we cannot have. But maybe it's possible and I just don't know how. If you are referring to "persistent overlays", I think I tried that but it requires root.

Ah, I see! I'm not into the whole LUMI-terminology.

Sounds like getting started with a simple dev-container, and then branching out to separate containers if required, is the way to go. We can always use a multi-stage build Dockerfile to make them composable.

rlrs:
Singularity can't even run Docker containers - it's another format. But let's get this merged and we can look at it later, seems like a separate issue that we don't need to solve now.

rlrs (approved):

PR #186, opened 2023-11-08
Dolma taggers

peterbjorgensen:
This adds dolma as a dependency and adds some custom taggers for Scandinavian languages.
The taggers are supposed to be used from the dolma command line tool.
The ability to use custom taggers from dolma command line tool is currently only available in the git repo, but will be included in the next release.

For example, we can tag mC4 dataset like this:
`PYTHONPATH=/path/to/src/dfm/common/data_cleaning/dolma_taggers dolma tag --documents /path/to/mC4_da/documents --experiment mytagsname --tagger_modules language_scandi perplexity --taggers cld2_scandi_paragraph_with_doc_score ft_lang_id_scandi_paragraph_with_doc_score char_length_with_paragraphs_v1 whitespace_tokenizer_with_paragraphs_v1 ccnet_perplexity_paragraph_w_doc_da --processes 10`


peterbjorgensen:
I had some problem with strict pyright not liking the unknown data types from the external libraries `pycld2`, `kenlm` and `blingfire`. I tried to generate stubs using pyright, but it just generated empty files. I guess it's possible to write the type files manually, but seems like a lot of work for a small benefit.

pyproject.toml
```diff
@@ -16,7 +16,13 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 
-dependencies = ["pydantic==1.8.2"]
+dependencies = [
+  "pydantic>=2.4.2", # dolma does not work with very old versions of pydantic
+  "dolma@git+https://github.com/allenai/dolma.git", # Install from git until a 0.9.2 package is released
+  "kenlm", # Used for perplexity tagging
+  "blingfire", # Used for perplexity tagging
+  "requests",
```
KennethEnevoldsen, line 24: Set a minimum version on these

pyproject.toml
```diff
@@ -16,7 +16,13 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 
-dependencies = ["pydantic==1.8.2"]
+dependencies = [
+  "pydantic>=2.4.2", # dolma does not work with very old versions of pydantic
+  "dolma@git+https://github.com/allenai/dolma.git", # Install from git until a 0.9.2 package is released
```
KennethEnevoldsen, line 21: Have you tested that this works? I feel like pyproject.toml reject links in the dependencies due to security. If it works feel free to resolve this comment

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
```diff
@@ -0,0 +1,202 @@
...
+    "DANISH": "da",
+    "SWEDISH": "sv",
+    "NORWEGIAN": "no",
+    "ICELANDIC": "is",
+    "FAROESE": "fo",  # Note that FAROESE is not supported by cld2
+}
+
+
+@TaggerRegistry.add("cld2_scandi_doc")
+class Cld2LanguageFilterScandi(BaseTagger):
```
KennethEnevoldsen, line 28: A small docstring to each of these would be great

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
```diff
@@ -0,0 +1,202 @@
...
+                    type=f"not_{lang_code}",
+                    score=1.0 - score,
+                )
+                spans.extend((positive_span, negative_span))
+        return DocResult(doc=doc, spans=spans)
+
+
+@TaggerRegistry.add("ft_lang_id_scandi_doc")
+class FastTextScandiLanguageDocumentTagger(BaseFastTextTagger):
+    MODEL_PATH = "https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.bin"
```
KennethEnevoldsen, line 111: Can we add a reference to the models (potentially in the docstring), would love to check the performance metrics.

KennethEnevoldsen (commented):
Looks good. I would love a few references on the models, but that is mostly to evaluate the taggers. 

I am unsure what the trade-offs are with the type stubs @MartinBernstorff do you have experience with those?

peterbjorgensen:
> Looks good. I would love a few references on the models, but that is mostly to evaluate the taggers.
> 
> I am unsure what the trade-offs are with the type stubs @MartinBernstorff do you have experience with those?

I have manually filled in the type information in the type stubs for the parts of the library that I have used, but left the other parts untyped, because it is too much work to cover the full library.

pyproject.toml
peterbjorgensen: Yes, it works. And yes, it is not very safe. It's temporary until a new package has been published.

pyproject.toml
MartinBernstorff: Perhaps a pinned version instead? Since DFM is more of an application than a library, and pinned versions maximise reproducibility :-)

MartinBernstorff:
> > Looks good. I would love a few references on the models, but that is mostly to evaluate the taggers.
> > I am unsure what the trade-offs are with the type stubs @MartinBernstorff do you have experience with those?
> 
> I have manually filled in the type information in the type stubs for the parts of the library that I have used, but left the other parts untyped, because it is too much work to cover the full library.

Absolutely! Only stub the part of the library we care about, if that. Nice work. Reviewing now 👍

pyproject.toml
MartinBernstorff, line 21: We can pin to a specific hash for maximum security. Alternatively, we probably want an issue so we don't forget this.

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
```diff
@@ -0,0 +1,202 @@
...
+    "SWEDISH": "sv",
+    "NORWEGIAN": "no",
+    "ICELANDIC": "is",
+    "FAROESE": "fo",  # Note that FAROESE is not supported by cld2
+}
+
+
+@TaggerRegistry.add("cld2_scandi_doc")
+class Cld2LanguageFilterScandi(BaseTagger):
+    RE_BAD_CHARS = regex.compile(r"[\p{Cc}\p{Cs}]+")
```
MartinBernstorff, line 29:
```suggestion
    RE_BAD_CHARS = regex.compile(r"[\p{Cc}\p{Cs}]+")
    # ﻿\p{Cc}: Matches a Unicode character with the "Other, Control" property, representing non-printing characters like newline, horizontal tab, escape, etc.
    # \p{Cs}: Matches a Unicode character with the "Other, Surrogate" property, which is used for representing characters from supplementary planes in UTF-16 (surrogate pairs).
```

MartinBernstorff: Do we want an `|` in between those two, or no?

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
```diff
@@ -0,0 +1,202 @@
...
+        return self.RE_BAD_CHARS.sub("", text)
+
+    def _to_ascii_input(self, text: str) -> str:
+        return anyascii(text)
+
+    def _identity_fn(self, text: str) -> str:
+        return text
+
+    def _predict_text(self, text: str) -> dict[str, float]:
+        is_reliable = False
```
MartinBernstorff, line 41: A comment here would be awesome. What does 'reliable' mean?

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
```diff
@@ -0,0 +1,202 @@
...
+        details: Iterable[tuple[str, str, int, float]] = []
+        for fn in (self._identity_fn, self._to_ascii_input, self._sanitize_input):
```
MartinBernstorff, line 43: Is there no case where `self._to_ascii_input(self._sanitize_input(text))` would work, but where `self._to_ascii_input` would not? 

If this works though, fuck it! ;-)

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
MartinBernstorff, line 28:
```suggestion
class Cld2LanguageClassifier(BaseTagger):
```

Perhaps? Since it tags, rather than filters.

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
```diff
@@ -0,0 +1,202 @@
...
+                score=doc_not_lang_score,
+            ),
+        )
+        result.spans.extend(doc_level)
+    return result
+
+
+# Composite tagger that provides both paragraph and doc scores
+@TaggerRegistry.add("cld2_scandi_paragraph_with_doc_score")
+class Cld2LanguageFilterParagraphWithDocScoreTaggerScandi(
```
MartinBernstorff, line 185: Same point re: naming as filter or not.

src/dfm/common/data_cleaning/dolma_taggers/perplexity.py
```diff
@@ -0,0 +1,180 @@
...
+}
+
+
+def _get_ccnet_pretrained_lm(lang: str) -> Path:
+    # Download pretrained model and save to the data folder
+    url = f"http://dl.fbaipublicfiles.com/cc_net/lm/{lang}.arpa.bin"
+    data_folder = Path("data_lm")
+
+    if not Path.exists(data_folder):
+        Path.mkdir(data_folder, parents=True)
```
MartinBernstorff, line 79:
```suggestion
    Path.mkdir(data_folder, parents=True, exist_ok=True)
```

Just a small tip 👍

src/dfm/common/data_cleaning/dolma_taggers/perplexity.py
```diff
@@ -0,0 +1,180 @@
...
+        f"CCNetPerplexity{lang}",
+        (PerplexityBaseTagger,),
+        {
+            "__init__": __init__,
+            "predict": predict,
+        },
+    )
+    # Add the class decorator explicitly to add the tagger to the registry
+    cls = TaggerRegistry.add(f"ccnet_perplexity_paragraph_w_doc_{lang}")(cls)
+    return cls
```
MartinBernstorff, line 176: I'm insufficiently proficient in how the registry works, but I assume you've tested this 👍

src/dfm/common/data_cleaning/text_normalizer.py
```diff
@@ -0,0 +1,200 @@
...
+        .replace("«", '"')
+        .replace(" » ", '" ')
+        .replace(" »", '"')
+        .replace("»", '"')
+        # handle pseudo-spaces
+        .replace(" %", "%")
+        .replace("nº ", "nº ")
+        .replace(" :", ":")
+        .replace(" ºC", " ºC")
+        .replace(" cm", " cm")
```
MartinBernstorff, line 127: Equal? 
image

I copy/pasted.

MartinBernstorff (approved):
Looks good! The normalisation is the kind of thing I'd personally write tests for, because I have a hard time knowing whether it is 'good enough' for our use-case. Might just mean that I lack context, though 👍  

I'd also remove the control-flow branches that are unrelated to scandi languages. 

The type-stubs are nice! My understanding is that we want as few as possible, though, so we only cover the surface area of the libraries that we need, but no more. This keeps maintenance low in case we need to update the packages in the future. I couldn't quite see if this is currently the case, or we kept as much as possible of pyright's output.

Looks awesome in general! Good work 👍

src/dfm/common/data_cleaning/text_normalizer.py
MartinBernstorff: Makes me less confident in the normalizer in general.

pyproject.toml
MartinBernstorff, line 21:
```suggestion
  "dolma@git+https://github.com/allenai/dolma.git@5a010a2685914b1db7744426abfb4b9ece52da95", # Install from git until a 0.9.2 package is released
```

src/dfm/common/data_cleaning/text_normalizer.py
peterbjorgensen: I have put the `text_normalizer.py` as a separate module, but the reason why I use exactly this one, is because it's the same one as CCNet uses and I use their model for perplexity tagging. I agree that it's not the nicest looking normalizer and maybe not something we want to use in general. Maybe I should rename it into `ccnet_text_normalier.py` to make it clear that it is just part of the ccnet perplexity pipeline.

src/dfm/common/data_cleaning/dolma_taggers/perplexity.py
peterbjorgensen: Yes, but it's a good point that I should add some tests that at least sanity checks the taggers.

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
peterbjorgensen: The square bracket expression `[\p{Cc}\p{Cs}]` is identical to `(\p{Cc}|\p{Cs})` and with the `+` it should match any sequence of 1 or more control characters.

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
peterbjorgensen: I have added docstrings to each class in a new commit.

pyproject.toml
peterbjorgensen: I see your point. I prefer minimum versions because it's much less strict and I think we should try to be compatible with the latest releases of libraries as long as the DFM project is maintained.

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
peterbjorgensen: Yes, my tagger naming scheme was in general very inconsistent. Will fix that.

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
peterbjorgensen: Added some more comments, but in this particular case it just means that the prediction was "high confidence", which doesn't tell me anything :)

src/dfm/common/data_cleaning/dolma_taggers/language_scandi.py
peterbjorgensen: It was able to run through mC4 Danish without problems, so I think it works.

PR #193, opened 2023-11-28
update dolma dependency to newest version, contains fix for dolma stat command

peterbjorgensen:
Also add the optional dependencies for pii (personally identifiable information) tagger and code tagger.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ requires-python = ">=3.10"
 dependencies = [
   "pydantic>=2.4.2", # dolma does not work with very old versions of pydantic
   "datasets>=2.4.0",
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen (approved):

KennethEnevoldsen:
Looks fine. Are the PII filters relevant for Danish?

peterbjorgensen:
> Looks fine. Are the PII filters relevant for Danish?

I think they use the PII filters built-in to Microsoft Presidio, but it only support English out of the box. In principle it can be extended to other languages, but maybe it's the same as starting from scratch https://microsoft.github.io/presidio/analyzer/languages/

KennethEnevoldsen:
Right, let us keep it an (assuming we add English.

PR #212, opened 2024-01-15
docs: Add munin to models page

saattrupdan:
This adds our Munin-7b-alpha model to the "Models" page on the website.

Also updates the Hugging Face organisation link on the "Models" page.


docs/models.md
```diff
@@ -1,16 +1,19 @@
 
 
-This section gives an overview of the models available through the DFM project. The models are available through the [Huggingface model hub](https://huggingface.co/models?filter=chcaa). To avoid duplicating information surrounding the models and the information regarding the models are available at the models model sheet.
+This section gives an overview of the models available through the DFM project. The
+models are available through the [Huggingface model
+hub](https://huggingface.co/danish-foundation-models). To avoid duplicating information
```
rlrs, line 5: Only the Munin model is available here right now. Should we change that?

KennethEnevoldsen: (we are also discussing the name change over on slack so will do the rename and move once)

KennethEnevoldsen (approved):
Looks good! We could create a collection of models at Danish Foundation Models with all the our models as well (though that would just be the org), but it might be worth excluding test models from that list. Not sure it is worth maintaining though.

PR #217, opened 2024-01-24
chore: Add CNAME

saattrupdan:
This supposedly fixes the issue with our disappearing custom domain after each commit:

https://github.com/mkdocs/mkdocs/issues/1257

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-www.foundationmodels.dk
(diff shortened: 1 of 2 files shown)


CNAME
KennethEnevoldsen, line 1: I don't believe this one is needed.

KennethEnevoldsen (approved):

CNAME
saattrupdan: Good point, removed!

PR #218, opened 2024-01-24
docs: Format "The Team" section

saattrupdan:
The "The Team" section was formatted a bit weird. It previously looked like this:

Screenshot 2024-01-24 at 16 39 58

With this change it now looks like this:

Screenshot 2024-01-24 at 16 40 13


KennethEnevoldsen (approved):
Thanks for the fix

PR #227, opened 2024-02-01
add script for converting scandi-reddit from HuggingFace to jsonl.gz

peterbjorgensen:


rlrs:
Can you just fix the path here?

PR #232, opened 2024-02-08
Scandeval on LUMI

rlrs:


rlrs:
Currently stuck due to vLLM/ROCm issues. I will probably work in another branch on using the built-in llm-foundry eval instead.

KennethEnevoldsen:
Will close this as it seems to have gotten stale

PR #235, opened 2024-02-09
219 reconsider ci settings

rlrs:
Just dumping some issues I have with our CI setup right now.

- I don't think the stale bot adds any value. It just adds bureaucracy and off-topic messages to issues and PRs. I would just remove it completely.
- CI pipelines run on every push to PRs and sends lots of emails about failures etc. Could we run this on demand instead?


rlrs:
Wonder if this command works now...
/build

rlrs:
Test

rlrs:
Okay so apparently you can't test these CI features on a PR. 
> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

In other words, we have to merge this to even test it.

saattrupdan:
@rlrs I have no idea what's happening here - can you add a PR description please? 😅

.github/workflows/stalebot.yml
KennethEnevoldsen, line 1: I am all for changing these thresholds, but I am unsure if I would delete it. Would love to hear what you think

.github/workflows/lint-and-build-and-test.yml
```diff
@@ -1,13 +1,14 @@
...
   push:
     branches:
       - main
-
+  issue_comment:
+    types: [created, edited]
 
 jobs:
   build-and-test:
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main' || github.event_name == 'issue_comment' && contains(github.event.comment.body, '/build')
```
KennethEnevoldsen, line 11: Is there any reason why we wouldn't just run it on each commit?

KennethEnevoldsen (changes requested):
Completely agree with @saattrupdan here

rlrs:
> @rlrs I have no idea what's happening here - can you add a PR description please? 😅

The linked issue has the explanation. 

> Is there any reason why we wouldn't just run it on each commit?

I guess this is a matter of taste, but I don't like being told on every commit that something is failing. I even get an email every time it fails. Why not just run it when needed?

saattrupdan:
> > @rlrs I have no idea what's happening here - can you add a PR description please? 😅
> 
> The linked issue has the explanation.
> 
> > Is there any reason why we wouldn't just run it on each commit?
> 
> I guess this is a matter of taste, but I don't like being told on every commit that something is failing. I even get an email every time it fails. Why not just run it when needed?

You can disable all the emails. And I think you can set it up in the repo settings whether the checks should be run automatically or not, but I might be wrong here

rlrs:
But I don't want to disable all the emails. I like getting an email about a workflow failing, as long as the workflow runs when I want it instead of every commit.

.github/workflows/stalebot.yml
rlrs: I'm not sure why we even need it? We're not overrun with issues that will never get looked at, which I think is the purpose of the bot. Currently all of our issues simply have no-stale because no one wants to deal with the bot.

saattrupdan:
> But I don't want to disable all the emails. I like getting an email about a workflow failing, as long as the workflow runs when I want it instead of every commit.

It seems as if this is very specific to your setup, and I fear that many PRs will be flooded with "/build" comments.

As a compromise, how about that we add a "manual-ci" label that we can add to our PRs, and the automatic triggering of the CI only happens when this label is attached to the PR in question?

rlrs:
I don't think it's fair to call the default setup that we're using "very specific" to me. But sure, if we do want to run the full workflow on every commit I can just turn off Actions notifications.

However if we do want such a strict approach, I think we should put more effort into having CI that is high quality and efficient so that it is easy to not get failing flows. Right now it just fails and it will keep failing on all commits - and we're not even running any actual tests, just (trying to) build the devcontainer over and over.

As for the stalebot, are there any arguments for it?

saattrupdan:
> I don't think it's fair to call the default setup that we're using "very specific" to me. But sure, if we do want to run the full workflow on every commit I can just turn off Actions notifications.

But, as far as I can see, the email notifications settings are on your personal Github profile settings, and are not currently set in the repo. I at least don't get any notifications at all, and I haven't changed anything. In my profile settings I see the following:

Screenshot 2024-02-12 at 13 06 32

> However if we do want such a strict approach, I think we should put more effort into having CI that is high quality and efficient so that it is easy to not get failing flows. Right now it just fails and it will keep failing on all commits - and we're not even running any actual tests, just (trying to) build the devcontainer over and over.

I completely agree that we need proper tests, that are not failing unless someone made a mistake. We can start by trying to fix the devcontainer Github action?

> As for the stalebot, are there any arguments for it?

No strong opinions about that one, I don't mind if we remove it 🙂

rlrs:
> But, as far as I can see, the email notifications settings are on your personal Github profile settings

Yeah exactly, you changed the default on your profile. I'm gonna do the same, but it's not default. And the notifications are mostly annoying because our CI is broken I guess.

.github/workflows/stalebot.yml
KennethEnevoldsen: Fair points - deleting it is probably fine for now

KennethEnevoldsen:
It seems like more of an issue of fixing the CI. I would just delete the dev container CI for now (along with outdated commands in the makefile).

saattrupdan:
> > But, as far as I can see, the email notifications settings are on your personal Github profile settings
> 
> Yeah exactly, you changed the default on your profile. I'm gonna do the same, but it's not default. And the notifications are mostly annoying because our CI is broken I guess.

Ah, that's what you meant by default - I thought you meant repo settings. My bad. But would the solution with the `manual-ci` label work for you in that case?

rlrs:
Well I removed the changes to the container workflow. I'm still suggesting to remove stalebot because I don't see why we bother.

saattrupdan:
> Well I removed the changes to the container workflow. I'm still suggesting to remove stalebot because I don't see why we bother.

Maybe we should just remove the `lint-and-build-and-test.yml` file as well, as it seems to only build a dev container atm?

Then we can set up a separate PR with proper CI, that actually works, as well as a `manual-ci` label.

saattrupdan (approved):
LGTM

PR #236, opened 2024-02-12
Fix submodule

rlrs:
The llm-foundry submodule was not set up correctly.


rlrs:
I don't think anyone could object to this, so I'm just gonna merge it.

PR #241, opened 2024-02-16
Url tagger

peterbjorgensen:
This adds an URL tagger based on the categories in the UT1 banlists.


KennethEnevoldsen (approved):
Looks good. Are actually applying this filter though?

peterbjorgensen:
We will apply these filters on future datasets, but was not used for 2024v1

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        'cuties',
+        'fap',
+        'fuck',
+        'gay',
+        'granny',
+        'horny',
+        'jerk',
+        'jizz',
+        'mature',
+        'moms',
```
KennethEnevoldsen, line 83: Seems problematic. Would remove

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        'porn',
+        'pussy',
+        'sex',
+        'shemale',
+        'spicytranny',
+        'tits',
+        'tube',
```
KennethEnevoldsen, line 90: YouTube? Seems problematic would remove

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        'gay',
+        'granny',
+        'horny',
+        'jerk',
+        'jizz',
+        'mature',
+        'moms',
+        'porn',
+        'pussy',
+        'sex',
```
KennethEnevoldsen, line 86: sexogsamfund. Seems problematic would remove.

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        'cum',
+        'cuties',
+        'fap',
+        'fuck',
+        'gay',
+        'granny',
+        'horny',
+        'jerk',
+        'jizz',
+        'mature',
```
KennethEnevoldsen, line 82: I am unsure about this would, but could see sites using this which is not porn

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        ".org",
+        )
+    WORD_BANLIST = (
+        'bbw',
+        'beauties',
+        'butt',
+        'butt',
+        'cock',
+        'cum',
+        'cuties',
```
KennethEnevoldsen, line 74: problematic?

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+    ALLOWLIST_TLD = (
+        ".dk",
+        ".se",
+        ".is",
+        ".no",
+        ".org",
+        )
+    WORD_BANLIST = (
+        'bbw',
+        'beauties',
```
KennethEnevoldsen, line 69: problematic?

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        'butt',
+        'butt',
+        'cock',
+        'cum',
+        'cuties',
+        'fap',
+        'fuck',
+        'gay',
```
KennethEnevoldsen, line 77: Besides a category it is also a sexuality, I would remove it.

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        'granny',
+        'horny',
+        'jerk',
```
KennethEnevoldsen, line 80: Can refer to "a jerk", "knee-jerk" etc. I would remove as well.

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+    return cls
+
+
+tagger_paths = os.environ.get("DOLMA_URL_CUSTOM_BLOCKLIST_PATH")
+
+if tagger_paths is not None:
+    for path in tagger_paths.split(","):
+        create_domain_blocklist_tagger(Path(path))
+
+@TaggerRegistry.add("adult_tld_v1")
```
KennethEnevoldsen, line 41: Generally have no issue running this filter to gain a list of domains, which we can the take the largest offenders from and then remove them or whitelist domains which contains these words but are not bad (I can do the validation if we have a public list). 

This can also help us establish a scale of the problem. Not sure if it is huge or not. I am even not sure it is something that we want to filter? I mean if someone want to write erotic literature then who am I to stop it.

My guess is however that there are smarter/more naive ways to do this (e.g. using removing sites which few people link to), which also filter poor quality.

data-processing/src/dfm_data/dolma_taggers/url.py
```diff
@@ -0,0 +1,106 @@
...
+        ".se",
+        ".is",
+        ".no",
+        ".org",
+        )
+    WORD_BANLIST = (
+        'bbw',
+        'beauties',
+        'butt',
+        'butt',
```
KennethEnevoldsen, line 71: duplicate

data-processing/src/dfm_data/dolma_taggers/url.py
KennethEnevoldsen: Generally female adjectives seems problematic

KennethEnevoldsen (commented):
Added a few comments

data-processing/src/dfm_data/dolma_taggers/url.py
peterbjorgensen: I made up this list after looking at the top domains in our dataset after filtering with the already established domain blockers built in to dolma. All the domains I saw were `.com` domains and I did not see any high-quality Danish sources that did not have `.dk` og `.org` as top-level domain. I think we could delete all `.com` domains without losing much quality.

PR #245, opened 2024-02-23
Add scripts for converting memo

TTTTao725:
Prerequisite: # Make sure you have git-lfs installed (https://git-lfs.com)
git lfs install
git clone https://huggingface.co/datasets/MiMe-MeMo/Corpus-v1.1

### DIFF
--- a/data-processing/scripts/infomedia/convert_infomedia_to_jsonlgz.py
+++ b/data-processing/scripts/infomedia/convert_infomedia_to_jsonlgz.py
@@ -1,0 +1,134 @@
+'''
+Converting infomedia dataset.
+
+ndjson -> json.gz:
+
+{
+    "id": "...",             # MANDATORY: source-specific identifier
+    "text": "foo",           # MANDATORY: textual content of the document
+    "source": "...",         # MANDATORY: source of the data, such as peS2o, common-crawl, etc.
+    "added": "...",          # OPTIONAL: timestamp we acquired this data (time file was created), specified as YYYY-MM-DDTHH:MM:SS.TIMEZONE e.g 2021-04-13T12:52:46.000Z
+    "created": "..."         # OPTIONAL: timestamp when orig document was created (best-guess if not available), should be specified as a range; "YYYY-MM-DDTHH:MM:SS.TIMEZONE, YYYY-MM-DDTHH:MM:SS.TIMEZONE"
+    "metadata": {            # OPTIONAL: source-specific metadata
+										"sub-source": "...", # OPTIONAL: E.g. "newspaper_ocr"
+										...
+								}        
+}
+'''
+
+import os
+import re
+import json
+import gzip
+import uuid
+import tempfile
+from datetime import datetime, timedelta
+from tqdm import tqdm
+from multiprocessing import Pool, cpu_count
+
+def format_created_range(publish_date):
+    """Create a formatted string representing a time range starting from `publish_date`."""
+    start_date = datetime.strptime(publish_date, "%Y-%m-%dT%H:%M:%SZ")
+    end_date = start_date + timedelta(days=365)
+    return f"{start_date.strftime('%Y-%m-%dT%H:%M:%SZ')}, {end_date.strftime('%Y-%m-%dT%H:%M:%SZ')}"
+
+def remove_html_tags(text):
+    """Remove HTML tags from a string."""
+    html_tag_pattern = re.compile('<.*?>')
+    clean_text = re.sub(html_tag_pattern, '', text)
+    return clean_text
+
+def process_file(filepath):
+    """Process a single file and write its processed contents to a temporary file."""
+    # Note: writing to the same file for all workers might lead to some problem.
+    articles = []
+    # Unique id
+    temp_filename = f"/work/github/utilities/temp_files/temp_{uuid.uuid4().hex}.json"
+    try:
+        if filepath.endswith('.ndjson') and os.path.getsize(filepath) > 0:
+            with open(filepath, 'r', encoding='utf-8') as file, open(temp_filename, 'w', encoding='utf-8') as temp_file:
+                for line in file:
+                    original = json.loads(line)
+                    added = datetime.now().strftime('%Y-%m-%dT%H:%M:%S.000Z')
+                    # Extract the fields
+                    heading = original.get("Heading", "")
+                    sub_heading = original.get("SubHeading", "")
+                    paragraph = original.get("Paragraph", "")
+                    body_text = original.get("BodyText", "")
+                    publish_date = original.get("PublishDate", "")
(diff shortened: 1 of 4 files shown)


data-processing/scripts/memo/convert_memo_to_jsonlgz.py
```diff
@@ -0,0 +1,94 @@
...
+    "text": "foo",           # MANDATORY: textual content of the document
+    "source": "...",         # MANDATORY: source of the data, such as peS2o, common-crawl, etc.
+    "added": "...",          # OPTIONAL: timestamp we acquired this data (time file was created), specified as YYYY-MM-DDTHH:MM:SS.TIMEZONE e.g 2021-04-13T12:52:46.000Z
+    "created": "..."         # OPTIONAL: timestamp when orig document was created (best-guess if not available), should be specified as a range; "YYYY-MM-DDTHH:MM:SS.TIMEZONE, YYYY-MM-DDTHH:MM:SS.TIMEZONE"
+    "metadata": {            # OPTIONAL: source-specific metadata
+										"sub-source": "...", # OPTIONAL: E.g. "newspaper_ocr"
+										...
+								}        
+}
+'''
```
KennethEnevoldsen, line 24:
```suggestion
'''
.txt (text) & .csv(metadata) -> json.gz:

{
    "id": "...",             # MANDATORY: source-specific identifier
    "text": "foo",           # MANDATORY: textual content of the document
    "source": "...",         # MANDATORY: source of the data, such as peS2o, common-crawl, etc.
    "added": "...",          # OPTIONAL: timestamp we acquired this data (time file was created), specified as YYYY-MM-DDTHH:MM:SS.TIMEZONE e.g 2021-04-13T12:52:46.000Z
    "created": "..."         # OPTIONAL: timestamp when orig document was created (best-guess if not available), should be specified as a range; "YYYY-MM-DDTHH:MM:SS.TIMEZONE, YYYY-MM-DDTHH:MM:SS.TIMEZONE"
    "metadata": {            # OPTIONAL: source-specific metadata
										"sub-source": "...", # OPTIONAL: E.g. "newspaper_ocr"
										...
								}        
}
'''


import pandas as pd
import json
import gzip
import os
from tqdm import tqdm
import datetime
import nltk
from nltk.tokenize import word_tokenize
```

assuming it is a docstring I would also write it out a bit. I.e. "This script converts ..."

peterbjorgensen (changes requested):
Can you set the created to a range instead of unknown if the year is unknown? I guess we know the range with an accuracy of +/- 100 years or something.

Otherwise it looks good to me.

KennethEnevoldsen (changes requested):
There are a few changes I would add otherwise looks good.

TTTTao725:
Hi guys, I changed the commit to make it more readable, please check it out :)

KennethEnevoldsen (approved):
Looks good

TTTTao725:
Hi guys, please check out the scripts for infomedia (converting and filtering)

unfiltered, filtered datasets and the list of empty files in the original dataset, all of them are here:
dfm-data/pre-training/danews2.0

e -i HEAD~2
KennethEnevoldsen, line 1: seems like this file should be deleted.

data-processing/scripts/infomedia/convert_infomedia_to_jsonlgz.py
KennethEnevoldsen, line 1: Next time I would have done this in another PR, but no it is no biggie.

data-processing/scripts/infomedia/convert_infomedia_to_jsonlgz.py
```diff
@@ -0,0 +1,134 @@
...
+    """Create a formatted string representing a time range starting from `publish_date`."""
+    start_date = datetime.strptime(publish_date, "%Y-%m-%dT%H:%M:%SZ")
+    end_date = start_date + timedelta(days=365)
+    return f"{start_date.strftime('%Y-%m-%dT%H:%M:%SZ')}, {end_date.strftime('%Y-%m-%dT%H:%M:%SZ')}"
+
+def remove_html_tags(text):
+    """Remove HTML tags from a string."""
+    html_tag_pattern = re.compile('<.*?>')
+    clean_text = re.sub(html_tag_pattern, '', text)
+    return clean_text
```
KennethEnevoldsen, line 39: @jankounchained how did you do the html filtering when you did it?

KennethEnevoldsen (approved):
Looks good. Please remove the odd additional file.

KennethEnevoldsen:
@TTTTao725 what is the number of tokens without information in infomedia?

e -i HEAD~2
TTTTao725: 👌🏼

TTTTao725:
6578205491 / 6753608085 tokens after filtering

KennethEnevoldsen:
Why the slash?

TTTTao725:
that's just the number before filtering :)

KennethEnevoldsen:
so Infomedia is ~175m tokens

TTTTao725:
Sorry for the confusion, I like to write it like a processing bar 😹

TTTTao725:
Oh I forgot to say, it does not include titles, summaries, etc, just text. So basically I just sum over the 'WordCount' instead of using a tokenizer which would be super slow

TTTTao725:
So it's actually more than 6578205491 :)

KennethEnevoldsen:
Ah so information is ~175m tokens

TTTTao725:
I can't merge it Kenneth, merging is block
image

KennethEnevoldsen:
just check the box and merge

data-processing/scripts/infomedia/convert_infomedia_to_jsonlgz.py
jankounchained: @KennethEnevoldsen @TTTTao725 sorry missed this!
the regex pattern looks good.

if you want to be extra careful, then you control for two edge cases:
- add a multiline flag in case there is a long html tag between mulitple lines
- replace html tags with a single shitespace in case that this happens: `word<br /></p><p id=\"p1\">word`
- take care of excess whitespace with another function

so something like this:
```
def remove_html_tags(text: str) -> str:
    """Remove HTML tags from a string."""
    html_tag_pattern = re.compile('<.*?>', flags=re.MULTILINE)
    clean_text = re.sub(html_tag_pattern, " ", text)
    return clean_text

def remove_whitespace(text: str) -> str:
    """remove excess whitespace from text fields"""
    pat_ws = re.compile(pattern=r"\s\s+", flags=re.MULTILINE)
    clean_text = re.sub(html_tag_pattern, " ", text)
    return clean_text
```

PR #250, opened 2024-03-06
Convert colossal oscar da

peterbjorgensen:
Add colossal oscar dataset (da) part. The script can easily be extended to all the other languages.

### DIFF
--- a/data-processing/scripts/convert_colossal_oscar_10_to_jsonlgz.py
+++ b/data-processing/scripts/convert_colossal_oscar_10_to_jsonlgz.py
@@ -1,0 +1,127 @@
+"""
+Convert from collection of files in this format:
+
+{
+   "content":"English sentence\nphrase en français\n????????????", // (1)
+   "warc_headers":{ // (2)
+      "warc-identified-content-language":"fra,eng",
+      "warc-target-uri":"https://fr.wikipedia.org/wiki/...",
+      "warc-record-id":"<urn:uuid:29eaa920-d299-4b1d-b687-c72bd8d68116>",
+      "warc-type":"conversion",
+      "content-length":"35298", // (3)
+      "warc-refers-to":"<urn:uuid:39e42055-0d94-4e45-9c6c-9e7056635d64>",
+      "warc-block-digest":"sha1:WFH2A5WHCS2H365GIAFYQPI7UOAMFGHB", // (3)
+      "warc-date":"2022-11-26T09:45:47Z",
+      "content-type":"text/plain"
+   },
+   "metadata":{
+      "identification":{ // (4)
+         "label":"fr",
+         "prob":0.8938327
+      },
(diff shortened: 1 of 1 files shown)


data-processing/scripts/convert_colossal_oscar_10_to_jsonlgz.py
```diff
@@ -0,0 +1,130 @@
...
+                output_path = output_dir / ("__".join(rel_input_path_parts[:-1]) + "__" + 
+                    Path(Path(rel_input_path_parts[-1]).stem).stem + f"_{doc_count//DOCS_PER_FILE}.jsonl.gz"
+                )
+                if out_fh is not None:
+                    out_fh.close()
+                out_fh = gzip.open(output_path, "wt")
+            obj = json.loads(line)
+            new_obj: dict[str, Union[dict[str, Any], str]] = {
+                "id": str(obj["warc_headers"]["warc-record-id"]), # Copy metadata id to root
+                "text": obj["content"],
```
KennethEnevoldsen, line 94: Is content cleaned from HTML?

data-processing/scripts/convert_colossal_oscar_10_to_jsonlgz.py
```diff
@@ -0,0 +1,130 @@
...
+    assert (
+        len(sys.argv) == 3
+    ), "Usage: convert_colossal_oscar_10_to_jsonlgz.py colossal_oscar_root output_directory"
+
+    oscar_root_dir = Path(sys.argv[1])
+    output_dir = Path(sys.argv[-1])
+
+    lang_codes = ["da", "sv", "no", "nn", "is"]
+    for lang in lang_codes:
+        input_files = oscar_root_dir.glob(f"./**/{lang}_meta/*.jsonl.zst")
```
KennethEnevoldsen, line 118: zst? Shouldn't we keep it all as json.gz?

KennethEnevoldsen (commented):
I assume there is a smart reason for choosing the format, but it would be nice to know.

@jankounchained you might be interested in this as well.

data-processing/scripts/convert_colossal_oscar_10_to_jsonlgz.py
peterbjorgensen: They use ungoliant which uses the WET files that are roughly cleaned from HTML, but it's probably full of noise. https://huggingface.co/datasets/oscar-corpus/colossal-oscar-1.0/blob/main/README.md

data-processing/scripts/convert_colossal_oscar_10_to_jsonlgz.py
peterbjorgensen: zst is only for the input (raw) dataset files. The output of the script is saved in jsonl.gz

KennethEnevoldsen (approved):

PR #251, opened 2024-03-11
Dolma url count

peterbjorgensen:
Add a simple (single proces) and a complicated (multiprocessing) script to count the domains in a dataset.


data-processing/scripts/dolma_url_count.py
```diff
@@ -0,0 +1,62 @@
...
+from typing import Any, Optional
+from pathlib import Path
+import sys
+
+import smart_open
+import msgspec
+from msgspec import Struct
+import urllib3.util
+
+class DocumentWithMetadata(Struct):
```
KennethEnevoldsen, line 11: Oh didn't know this was a thing!

KennethEnevoldsen (approved):
Not much to add here. I assume you have already run it. Is the files located on UCloud?

peterbjorgensen:
The domain list is on ucloud here
`dfm-data/domain-list`

KennethEnevoldsen:
@peterbjorgensen should we have the domain lists on Github (to allow users to remove/add their domains from the list)

PR #259, opened 2024-04-10
add gopher tagger with scandi stop words instead of english

peterbjorgensen:


rlrs (approved):
No wonder this takes a long time, it looks like an extremely slow implementation.

PR #264, opened 2024-04-22
NCC loader

jankounchained:
Added a minimally cleaned Norwegian Colossal Corpus.  
The processed files are in `dfm-data/pre-training/ncc`.  

**duplicates**
The data already comes with paragraph-level deduplication according to the [paper](https://aclanthology.org/2021.nodalida-main.3.pdf). 
I'm still running our deduplication as a formality so that the `bff_duplicate_paragraph_spans` folder isn't empty.

**language filtering**
A language tag from fasttext exists in the metadata. 
I decided not to use it in the end, because the share of tokens in non-nordic languages seems quite small to me.
@KennethEnevoldsen do you agree?

Non-Germanic languages make up about 1.3% of the corpus.  
Non-North-Germanic languages are 8.6% of the tokens.  
This difference is mainly due to English, 6.7% tokens.
Danish has 13.7% tokens. 

**ocr**
A big part of the corpus comes from OCR & the authors haven't done much about OCR errors.  
Quote: "We have not seen any indication that the OCR errors negatively impacted the performance."  
But, they also indicate that OCR quality is time-dependent, better or worse depending on a document's age. 

**licensing**
online newspapers (7.6% tokens) have a non-commercial clause (CC BY-NC 2.0). 
the wikipedia dump (2% of tokens) has a share-alike clause (CC BY-SA 3.0).
depending on where the project goes, these may need to be dropped?
rest of the data is released under very permissive licenses.

### DIFF
--- a/data-processing/scripts/convert_ncc_to_jsonlgz.py
+++ b/data-processing/scripts/convert_ncc_to_jsonlgz.py
@@ -1,0 +1,109 @@
+"""
+Download  The Norwegian Colossal Corpus, convert to jsonl.gz with each document following the format:
+
+{
+    "id": "...",             # MANDATORY: source-specific identifier
+    "text": "foo",           # MANDATORY: textual content of the document
+    "source": "...",         # MANDATORY: source of the data, such as peS2o, common-crawl, etc.
+    "added": "...",          # OPTIONAL: timestamp ai2 acquired this data
+    "created": "..."         # OPTIONAL: timestamp when orig document was created (best-guess if not available)
+    "metadata": {...}        # OPTIONAL: source-specific metadata
+}
+"""
+
+import os
+import datetime
+from functools import partial
+from datasets import load_dataset, Dataset, IterableDataset
+
+
+EXPORT_PATH = "ncc_sample.jsonl.gz"
+date_added = datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%S.000Z")
+
+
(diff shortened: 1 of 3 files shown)


data-processing/scripts/convert_ncc_to_jsonlgz.py
```diff
@@ -0,0 +1,109 @@
...
+        "id": obs["id"],
+        "text": obs["text"],
+        "source": "NCC",
+        "added": date_added,
+        "created": f"{publish_year}-01-01T00:00:00.000Z",
+        "metadata": {
+            "doc_type": obs["doc_type"],
+            "lang_fasttext": obs["lang_fasttext"],
+            "lang_fasttext_conf": obs["lang_fasttext_conf"],
+        },
```
KennethEnevoldsen, line 63: would be great to have license here as well

KennethEnevoldsen (commented):
Only a minor comment.

Read through your message @jankounchained the decisions seem very reasonable.

data-processing/scripts/convert_ncc_to_jsonlgz.py
jankounchained: done

KennethEnevoldsen:
Looks good Jan - feel free to merge this in - will you run it on the whole NCC as well?

jankounchained:
already did:)
if you mean the train split of the dataset on huggingface
`dfm-data/pre-training/ncc/documents/ncc.jsonlgz`

KennethEnevoldsen:
Wonderful!

PR #271, opened 2024-05-13
docs: Added user-friendly tutorial for qlora

KennethEnevoldsen:


saattrupdan (commented):
Don't really understand. It seems like the `finetune_qlora` notebook is meant to replace the `finetune` notebook, as both are QLoRa finetuning decoder models, but the new one is more user friendly. What's the point of the old one?

KennethEnevoldsen:
Well since it is a blog post I don't want to break any links. For practitioners the old one might also be more application-friendly(more memory efficient with less boilerplate and more config objects). I think both are relevant.

saattrupdan:
> Well since it is a blog post I don't want to break any links. For practitioners the old one might also be more application-friendly(more memory efficient with less boilerplate and more config objects). I think both are relevant.

But it's not a blog post, it's a new notebook. I'm all for writing new blog posts, but shouldn't we just update the existing notebook and make a new blog post about it? No links are broken then. Having two notebooks doing exactly the same thing just seems super confusing.

KennethEnevoldsen:
blog post == notebook (at least in our case)

We can def. also update the previous one. I would be totally be up for that as well though I would probably just replace it with the current one. There are however a few cases the previous notebook cover which this one does not (export to other formats)

KennethEnevoldsen:
I am gonna merge this in

PR #278, opened 2024-05-27
Fix wrong license danews2.0.md

KennethEnevoldsen:
@TTTTao725 it seems like the license is wrong on DaNews 2.0, not sure where that happened.

### DIFF
--- a/docs/datasheets/danews2.0.md
+++ b/docs/datasheets/danews2.0.md
@@ -2,8 +2,8 @@
 pretty_name:  Danish news 2.0
 language:
   - da
-license: cc0-1.0
-license_name: Creative Commons Zero v1.0 Universal
+license: other
+license_name: agreement
 size_categories:
   - 10M<n<100M
 task_categories:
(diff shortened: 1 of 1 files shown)


docs/datasheets/danews2.0.md
```diff
@@ -44,12 +44,5 @@ An example from the dataset looks as follows.
...
-<details>
-<summary>Creative Commons Zero v1.0 Universal</summary>
-<p>
-Creative Commons Legal Code
-
-CC0 1.0 Universal
-</p>
-</details>
+## License Information
+Agreement with individual newspapers
```
KennethEnevoldsen, line 48: We should probably specify this further. 

```suggestion
Agreement with individual newspapers. The agreement is for the Danish Foundation Models project and that models trained on the dataset can be published.
```

TTTTao725:
Great! Thanks!

TTTTao725 (approved):

PR #279, opened 2024-05-27
Add dataset guide

KennethEnevoldsen:
Dear Peters,

Added a guide for adding a dataset here. Let me know if anything is unclear.

### DIFF
--- a/docs/Adding_a_new_dataset
+++ b/docs/Adding_a_new_dataset
@@ -13,7 +13,9 @@ pre-training
 └── dataset_name
     │
     ├── documents
-    │   └── dataset_name.jsonl.gz
+    │   ├──part1.jsonl.gz
+    │   ├──part2.jsonl.gz
+    │   └── ...
     │
     └── attributes   # OPTIONAL: folder containing annotations from dataset cleaning
 ```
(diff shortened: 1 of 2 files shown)


docs/Adding_a_new_dataset
```diff
@@ -0,0 +1,40 @@
...
+   2) Languages (e.g. "da" for Danish)
+2) Add a dataset of the same name to the `/danish-foundation-models (193701)/dfm-data/pre-training/` on UCloud. Using the following folder structure:
+
+```
+pre-training
+│
+└── dataset_name
+    │
+    ├── documents
+    │   └── dataset_name.jsonl.gz
```
peterbjorgensen, line 16: The jsonl files do not need to have the same name as `dataset_name`, and this part of the documentation implies that it is a requirement. Also, it's recommended that the documents are split across multiple files, which makes parallel data processing easier (approximately 1M documents or around 1 GB per file is a sensible tradeoff), so I think it's better if the example shows `part1.jsonl.gz`, `part2.jsonl.gz` etc.

peterbjorgensen (changes requested):
Looks good to me, except for the minor comment.

docs/Adding_a_new_dataset
KennethEnevoldsen, line 16:
```suggestion
    │   ├──part1.jsonl.gz
    │   ├──part2.jsonl.gz
    │   └── ...
```

KennethEnevoldsen: @TTTTao725 just to make sure you see this

peterbjorgensen (approved):

PR #280, opened 2024-05-29
Add datatrove pipeline blocks

peterbjorgensen:
Add data pipeline blocks for datatrove

- PII formatter that also removes Danish CPR numbers
- Port/wrapper of dolma domain filters to datatrove
- adapter function to load dolma data format into datatrove


data-processing/src/dfm_data/datatrove_pipeline/url_filter.py
```diff
@@ -0,0 +1,437 @@
...
+from huggingface_hub import cached_assets_path
+from loguru import logger
+
+from datatrove.data import Document
+from datatrove.io import safely_create_file
+
+from datatrove.pipeline.writers.disk_base import DiskWriter
+from datatrove.pipeline.filters.base_filter import BaseFilter
+
+SOFT_DOMAIN_WORD_BANLIST_EXTRA = (
```
KennethEnevoldsen, line 17: is it used anywhere?

KennethEnevoldsen (approved):
Only one minor thing

data-processing/src/dfm_data/datatrove_pipeline/url_filter.py
peterbjorgensen: It's meant for being imported into a script and be used as keyword argument `soft_banned_words` of the datatrove URL filter, which will add these words to the standard list.

PR #287, opened 2024-06-05
datasheets for the four datasets from DBC D1G1TAL

peter-sk:


KennethEnevoldsen:
Looks great @peter-sk - I can't see the data on UCloud is it possible that I have missed something?

peter-sk:
You were too fast. Look again :-)

docs/datasheets/dbc-abstracts.md
```diff
@@ -0,0 +1,136 @@
+
```
KennethEnevoldsen, line 1: Seems like we are missing the frontmatter.

We are notably also missing license field in the datasheet. It should just describe what we can and can't do with the data if no license is specified.

KennethEnevoldsen (commented):
Ahh thanks see them now.

docs/datasheets/dbc-abstracts.md
peter-sk: What is the frontmatter?

peter-sk: Also, I am not sure how the datasheets should be formatted. I oriented myself according to daradio. But there seem to be vastly different datasheet styles used.

What is the preferred style? And how to put a license that says:

Data may be used to train on and model may be published openly. But data may not be shared with third parties.

KennethEnevoldsen: The frontmatter is a yaml segment at the beginning of a markdown separated by `---`, e.g.

```
---
{yaml}
---
{markdown}
```
In terms of content daradio is quite good, except for the frontmatter + license stuff. For an example of the frontmatter see e.g. [this](https://github.com/centre-for-humanities-computing/danish-foundation-models/blob/main/docs/datasheets/dagw-gutenberg.md)

the frontmatter is intended for the sheet to be machine readable.

peter-sk: Done!

docs/datasheets/dbc-abstracts.md
```diff
@@ -0,0 +1,153 @@
+---
+pretty_name: dbc-abstracts
+language:
+  - da
+  - en
+  - "no"
+  - sv
+  - ty
```
KennethEnevoldsen, line 8: ty?

KennethEnevoldsen (approved):
General looks fine, feel free to merge

docs/datasheets/dbc-abstracts.md
peter-sk: Oops. I de-danished the country code, relying on ISO-639-1 instead.

PR #317, opened 2025-03-24
Convert CI to UV

KennethEnevoldsen:
Convert CI to UV and ensure ci makefile consistency.

Also added UV lock, but shouldn't change much with locked dependencies

Also removed an outdated section from the documentation.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -21,6 +21,7 @@ jobs:
         uses: astral-sh/setup-uv@v4
         with:
           enable-cache: true
+          python-version: "3.10"
 
       - uses: actions/setup-python@v5
         with:
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -15,7 +15,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Install uv
         uses: astral-sh/setup-uv@v4
@@ -28,7 +28,7 @@ jobs:
           python-version: '3.10'
       
       - name: Dependencies
-        run: make install
+        run: uv sync
 
       - name: Build and Deploy
         if: github.event_name == 'push'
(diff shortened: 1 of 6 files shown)


.github/workflows/documentation.yml
```diff
@@ -14,21 +16,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
```
saattrupdan, line 18:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/documentation.yml
```diff
@@ -14,21 +16,23 @@ jobs:
...
-      - uses: actions/setup-python@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
         with:
-          python-version: "3.10"
-      - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV 
-      - uses: actions/cache@v3
+          enable-cache: true
```
saattrupdan, line 23:
```suggestion
          enable-cache: true
          python-version: "3.10"
```

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[project]
```
saattrupdan, line 1: Should we put authors in here? And license?

.github/workflows/documentation.yml
```diff
@@ -14,21 +16,23 @@ jobs:
...
-          key: mkdocs-material-${{ env.cache_id }}
-          path: .cache
-          restore-keys: |
-            mkdocs-material-
+          python-version: '3.10'
       
       - name: Dependencies
-        run: |
-          pip install -r docs/requirements.txt
+        run: make install
```
saattrupdan, line 30: I don't see this set up anywhere? Alternatively:

```suggestion
        run: uv sync
```

saattrupdan (changes requested):

pyproject.toml
KennethEnevoldsen: I don't want to maintain the author list

The license is fairly minor though

PR #318, opened 2025-05-19
Integrating DFM landing page into the site

peter-sk:


mkdocs.yml
```diff
@@ -1,8 +1,12 @@
 site_name: Danish Foundation Models
+site_url: https://peter-sk.github.io/site/docs/
+use_directory_urls: false
+
 docs_dir: "docs/"
-repo_url: https://github.com/centre-for-humanities-computing/danish-foundation-models
+repo_url: https://github.com/peter-sk/site
```
saattrupdan, line 6: Shouldn't this refer to this current repo instead of your private one?

mkdocs.yml
```diff
@@ -1,8 +1,12 @@
 site_name: Danish Foundation Models
+site_url: https://peter-sk.github.io/site/docs/
```
saattrupdan, line 2: Shouldn't this be foundationmodels.dk?

.github/workflows/documentation.yml
```diff
@@ -26,14 +25,35 @@ jobs:
...
+      - name: Copy root index.html and add .nojekyll
+        if: github.event_name == 'push'
+        run: |
+          cp index.html site/index.html
+          touch site/.nojekyll
+
+      - name: Clone landing page
+        if: github.event_name == 'push'
+        run: |
+          git clone https://github.com/schneiderkamplab/dfm-landing-page tmp-landing
```
saattrupdan, line 45: I guess this works, but wouldn't it be easier if you just moved the landing page inside this repo?

saattrupdan (changes requested):

mkdocs.yml
peter-sk: Sure. That's why it's still a draft. I needed to test it with the private fork.

mkdocs.yml
peter-sk: Let's try that.

.github/workflows/documentation.yml
peter-sk: Sure, that would be a possibility. The landing page also has the Flask backend code etc.
It will be a bit of a merge, but if you prefer this, we can, of course, do this.

KennethEnevoldsen (commented):
I would probably prefer keeping all the code in one place, but if not I would update the docs to clearly describe the workflow. 

A few things are still unclear to me:

1) How do we update the site? 
2) How will the site end up looking (the one you linked didn't seem like it used mkdocs). Adding a screenshot using this built would be great

.github/workflows/documentation.yml
peter-sk: The backend stores and delivers the dynamic content, of course.

peter-sk:
I have now adjusted the URIs according to a deployment.
The landing page will NOT be index.html to start with but landing.html such that we can test and refine the content without untoward attention.
Later, it will be a trivial swap of some paths to have the landing page be the default.

.github/workflows/documentation.yml
```diff
@@ -26,14 +25,33 @@ jobs:
...
           python-version: '3.10'
-      
+
       - name: Dependencies
         run: uv sync
 
-      - name: Build and Deploy
+      - name: Build MkDocs
+        if: github.event_name == 'push'
+        run: uv run mkdocs build
```
KennethEnevoldsen, line 34: this is the same as `make build-docs`

We ca just combine it with the PR Only

backend/requirements.txt
```diff
@@ -0,0 +1,4 @@
+flask
```
KennethEnevoldsen, line 1: move to pyproject.toml

KennethEnevoldsen (approved):
Sounds like we are good on a merge here, then we can polish the site afterwards

peter-sk:
Looks like all is good :-)

PR #320, opened 2025-05-21
Clean configuration and start of the backend

peter-sk:


dfm_backend/cli.py
```diff
@@ -0,0 +1,12 @@
...
+from pathlib import Path
+
+def run():
+    subprocess.run([
+        "gunicorn",
+        "dfm_backend:create_app()",
+        "-c",
+        str(Path(__file__).resolve().parent / "gunicorn.conf.py"),
+        *sys.argv[1:],
+    ])
```
saattrupdan, line 12: Since it seems like this is only here to be used with the `make start-backend` recipe, called through the `dfm-backend` package script, couldn't we simplify it by simply calling gunicorn directly in the make recipe instead of going via the package script?

dfm_backend/app.py
```diff
@@ -0,0 +1,322 @@
...
+
+def create_app(config: dict[str, Any] | str | Path | None = None) -> Flask:
+    if config is None:
+        config = getenv('APP_CONFIG', Path(__file__).resolve().parent / 'app.conf.py')
+    if isinstance(config, (str, Path)):
+        config_path = Path(config).absolute()
+        if not Path(config_path).exists():
+            raise FileNotFoundError(f"Config file {config_path} does not exist.")
+        config = {}
+        exec(open(config_path).read(), {}, config)
```
saattrupdan, line 21: Calling `exec` on arbitrary code seems like a potential security risk.

dfm_backend/app.py
```diff
@@ -0,0 +1,322 @@
...
+import uuid
+
+def create_app(config: dict[str, Any] | str | Path | None = None) -> Flask:
+    if config is None:
+        config = getenv('APP_CONFIG', Path(__file__).resolve().parent / 'app.conf.py')
+    if isinstance(config, (str, Path)):
+        config_path = Path(config).absolute()
+        if not Path(config_path).exists():
+            raise FileNotFoundError(f"Config file {config_path} does not exist.")
+        config = {}
```
saattrupdan, line 20: Generally, this way of fetching the configuration seems quite confusing to me. I get that the config variables are dynamic, so defining them in a separate Python module makes sense. But why not simply define a dictionary with all these values in `app.conf.py`, and just import that here with, e.g., `importlib`?

dfm_backend/app.py
```diff
@@ -0,0 +1,322 @@
...
+    mail = Mail(app)
+    CORS(app)
+
+    TOKENS = {}
+
+    def load_news():
+        if not Path(app.config['NEWS_FILE']).exists():
+            return []
+        with open(app.config['NEWS_FILE']) as f:
+            return [json.loads(line) for line in f]
```
saattrupdan, line 37: We will get empty lines in the JSONL file, and this just makes sure that we don't store those as news.

```suggestion
            return [json.loads(line) for line in f if line]
```

dfm_backend/app.py
```diff
@@ -0,0 +1,322 @@
...
+            
+    @app.route('/api/news', methods=['GET'])
+    def get_news():
+        return jsonify(load_news())
+
+    @app.route('/api/news', methods=['POST'])
+    def post_news():
+        auth = request.headers.get('Authorization', '')
+        token = auth.replace('Bearer ', '')
+        token_expiry = TOKENS.get(token, None)
```
saattrupdan, line 51: Nit: None is default.

```suggestion
        token_expiry = TOKENS.get(token)
```

dfm_backend/app.py
```diff
@@ -0,0 +1,322 @@
...
+        item = {
+            'id': str(uuid.uuid4()),
+            'title': data.get('title', '').strip(),
+            'date': data.get('date', '').strip(),
+            'content': data.get('content', '').strip()
+        }
+        if not item['title'] or not item['date'] or not item['content']:
+            return jsonify({'error': 'Missing fields'}), 400
+        save_news(item)
+        return jsonify(item), 201
```
saattrupdan, line 65: You can consider using the `HTTPStatus` from the built-in `http` package for the http status codes, to ease readability. For instance, 201 could be replaced by `HTTPStatus.CREATED` (which is just the integer 201 as well, so it's a drop-in replacement).

dfm_backend/cli.py
peter-sk: Like this, one can pip install and then just run dfm-backend. This seems a bit more clean but requires a Python wrapper.

dfm_backend/app.py
peter-sk: I am not sure how this constitutes a security risk. The user can already run the app startup script. This is not user-supplied but admin-supplied data.

dfm_backend/app.py
peter-sk: Yes, that would be possible. How is this different from exec? The code will be executed on import.
Also, we need to have the ability to override with app.conf.local.py for passwords etc.

dfm_backend/app.py
peter-sk: Done.

dfm_backend/app.py
saattrupdan: I guess it's maybe up to preference. Being explicit just seems a bit nicer than relying on "mysterious" side effects, that's all. Overriding will still be possible with imports, exactly as it is now.

dfm_backend/app.py
saattrupdan: Good point re admin-supplied. I'll resolve this comment and move the discussion to the other comment about using importlib instead of exec.

dfm_backend/cli.py
saattrupdan: Aha, I see. But if you pip install and run `dfm-backend`, it would still require you to be in a directory whose parent has a `gunicorn.conf.py` file, right? If so, maybe the path to that file should be a CLI argument to the `run` function?

dfm_backend/app.py
```diff
@@ -62,7 +63,7 @@ def post_news():
         if not item['title'] or not item['date'] or not item['content']:
             return jsonify({'error': 'Missing fields'}), 400
```
saattrupdan, line 64: Same thing for these other status codes as for 201 - would improve readability to replace them all with `HTTPStatus`'s.

dfm_backend/cli.py
peter-sk: no, the gunicorn.conf.py is installed in the package directory

dfm_backend/app.py
peter-sk: done

KennethEnevoldsen (approved):
Not a lot to add here.

A minor thing is that the requirement could be moved to pyproject.toml

PR #321, opened 2025-05-22
simpler and more integrated build with extended Makefile without weir…

peter-sk:
…d cli run.py


pyproject.toml
```diff
@@ -13,12 +13,13 @@ dependencies = [
     "flask-limiter==3.12",
     "flask-mail==0.10.0",
     "gunicorn==23.0.0",
+    "uv==0.7.6",
```
saattrupdan, line 16: This seems like an odd dependency to me - aren't we using `uv` as package manager? Is this just using `uv` to fix a version of `uv` for this project?

pyproject.toml
```diff
@@ -13,12 +13,13 @@ dependencies = [
...
 license-files = ["LICENSE"]
 
-[project.scripts]
-dfm-backend = "dfm_backend.cli:run"
-
 [tool.setuptools]
 packages = ["dfm_backend"]
+
+[tool.setuptools.data-files]
+"bin" = ["scripts/dfm-backend"]
```
saattrupdan, line 25: Haven't come across this before, what does it do? Does it include the directory in the Python package? And are the quotes around bin necessary?

scripts/dfm-backend
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+export GUNICORN_CONF=$(python -c $'import pathlib as p\nimport dfm_backend as d\nprint(p.Path(d.__file__).resolve().parent / "gunicorn.conf.py")')
```
saattrupdan, line 2: Is this used to get the path to the file inside the virtual environment? Seems quite hacky.

Makefile
```diff
@@ -8,6 +10,18 @@ view-docs:
...
+	@echo "--- 🏗️ Building site ---"
+	@echo "Builds the site and puts it in the 'site' folder"
+	mkdir -p site
+	touch site/.nojekyll
+	cp -rv landing/* site/
+
 start-backend:
 	@echo "--- 🚀 Starting backend ---"
-	dfm-backend
+	scripts/backend.sh
```
saattrupdan, line 27: Haven't seen this way of calling shell scripts in make recipes - we don't have to prepend a `sh ` to call it?

pyproject.toml
KennethEnevoldsen:
```suggestion
```

saattrupdan: uv is a replacement for pip, so it becomes a bit circular to install it with pip (but definitely possible).

The [recommended way](https://docs.astral.sh/uv/getting-started/installation/) to install uv is as follows:

```
curl -LsSf https://astral.sh/uv/install.sh | sh
```

pyproject.toml
peter-sk: It copies the script to the right bin directory (e.g., in a conda environment, env/NAME/bin; similar for other or no environments).
The "" around bin could be removed.

scripts/dfm-backend
peter-sk: do you have a better idea? i'm not sure how to do this.

Makefile
peter-sk: we are back to scripts/dfm-backend.
the file is executable (thanks to a shebang).

pyproject.toml
peter-sk: could we have that in the Makefile (who was spelling it "makefile" in the first place?)

scripts/dfm-backend
peter-sk: Kenneth didn't like the "Python file to run gunicorn to run Python" indirection. I'm more than happy to go back to that.

pyproject.toml
saattrupdan: Both makefile and Makefile are correct 🙂 

I usually have these make recipes in my repos, which also installs uv if it isn't already - you can use them as inspiration:

```make
install: ## Install dependencies
	@echo "Installing the 'X' project..."
	@$(MAKE) --quiet install-rust
	@$(MAKE) --quiet install-uv
	@$(MAKE) --quiet install-dependencies
	@echo "Installed the 'X' project."

install-rust:
	@if [ "$(shell which rustup)" = "" ]; then \
		curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y; \
		echo "Installed Rust."; \
	fi

install-uv:
	@if [ "$(shell which uv)" = "" ]; then \
		curl -LsSf https://astral.sh/uv/install.sh | sh; \
			echo "Installed uv."; \
		else \
			echo "Updating uv..."; \
		uv self update; \
	fi

install-dependencies:
	@uv python install 3.11
	@uv sync --all-extras --python 3.11
```

scripts/dfm-backend
saattrupdan: @KennethEnevoldsen

pyproject.toml
peter-sk: Done!

